PR #461, opened 2018-08-13
Add route docs in EN

HenryQW:
Remove most of the routes due to lack of multilingual support from the target data source


DIYgod:
Merged, thanks!

PR #482, opened 2018-08-16
Add appstore/price && Update appstore/update

HenryQW:


HenryQW:
Made the title to be more specific.
Get html instead of text so that the desctipion is readable.

HenryQW:
Add appstore/price for app price drop alert

PR #565, opened 2018-08-28
Update keyword.js

yuanotes:
add limits filter to weibo keyword source


yuanotes:
目前的关键字内容太水，没法看

yuanotes:
或者我加个粉丝数过滤

DerekTso:
@DIYgod @yuanotes when one weibo update comes in, make a delay, say, 10mins, then begin to fetch those info as a rss feed. is it ok to create a delay middleware for that case? just my assumption.

DIYgod:
内容太水属于数据源的问题，这样过滤意义不大

PR #623, opened 2018-09-02
Fix PKU EECS Notification Link return 404

dongdongdu:
1. fix item link 404 error 
2. add default title


DIYgod:
Merged, thanks!

PR #753, opened 2018-09-19
Add a thorough MIME image type check

HenryQW:
#555 
https://github.com/DIYgod/RSSHub/issues/555#issuecomment-422837421


DIYgod:
Merged, thanks!

PR #755, opened 2018-09-19
Correct the epoch timestamp

HenryQW:
Fix https://github.com/DIYgod/RSSHub/issues/673#issuecomment-422619248


DIYgod:
Merged, thanks!

PR #760, opened 2018-09-21
Remove deprecated npm package crypto

Maecenas:
```
npm WARN deprecated crypto@1.0.1: This package is no longer supported.
It's now a built-in Node module. If you've depended on crypto, you should switch to the one that's
built-in.
```


DIYgod:
Merged, thanks!

PR #907, opened 2018-10-16
add feed for aozora.gr.jp

sgqy:


sgqy:
There is a variable containing unicode string named `author`. It is encoded as `&x00000;`.
I tried some ways from stackoverflow.com but none works. :(

HenryQW:
> There is a variable containing unicode string named `author`. It is encoded as `&x00000;`.
> I tried some ways from stackoverflow.com but none works. :(

Try iconv-lite:
https://github.com/DIYgod/RSSHub/blob/143862e8bfcd9920251417c61630aff533ed062d/routes/tingshuitz/guangzhou.js#L14

PR #1056, opened 2018-11-02
fix pixiv datetime

sgqy:
* add author label in other 2 mode
* use real publish date
* remove ambiguous date string


DIYgod:
Merged, thanks!

PR #1094, opened 2018-11-09
Enhancement: Add support for private channel

### LINKED ISSUE #920: Telegram private channels
Well, there is no support for private channels, and its hard to understand why. The public URL can be set as `http://rsshub.app/telegram/channel/<id>`, id is the unique identifier of channels, i.e. -111111111.

xyqfer:
Closes #920


DIYgod:
Merged, thanks!

xamgore:
When will this be in production? :upside_down_face:

PR #1105, opened 2018-11-10
Translate info about telegram private channels

xamgore:
This is a logical continuation of the pull request about [support for private channels](https://github.com/DIYgod/RSSHub/pull/1094).


DIYgod:
Merged, thanks!

PR #1432, opened 2019-01-18
sspai matrix

feigaoxyz:
Add support for sspai matrix column [^1].

[^1] https://sspai.com/matrix


feigaoxyz:
@DIYgod please review the update and advice.

PR #1450, opened 2019-01-23
Show fulltext description for Sspai Matrix

feigaoxyz:
Follow up on #1432 to show full text description instead of summary.


DIYgod:
Merged, thanks!

Issue #353, opened 2018-07-10
Anitama wanted


benpigchu:
http://anitama.cn

benpigchu:
@ciskonc this one has a ~1day delay, but still usable.

PR #1770, opened 2019-03-17
add anitama

ranpox:
#353 the official RSS feed(https://app.anitama.net/rss/google) only shows titles with one day delay.


DIYgod:
Merged, thanks!

PR #1969, opened 2019-04-21
Add GitHub branches

max-arnold:
Turn an url like https://api.github.com/repos/octocat/Hello-World/branches into RSS feed

### DIFF
--- a/lib/routes/github/branches.js
+++ b/lib/routes/github/branches.js
@@ -1,0 +1,25 @@
+const axios = require('../../utils/axios');
+
+module.exports = async (ctx) => {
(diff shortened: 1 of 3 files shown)


lib/routes/github/branches.js
```diff
@@ -0,0 +1,26 @@
+const axios = require('../../utils/axios');
+const config = require('../../config');
```
DIYgod: 多余代码

max-arnold:
@DIYgod I removed the `const config` line.

DIYgod:
Merged, thanks!

PR #2107, opened 2019-05-11
fix: idownloadblog media lazy loading

HenryQW:


lib/routes/idownloadblog/index.js
```diff
@@ -31,7 +22,7 @@ module.exports = async (ctx) => {
         feed.items.map(async (item) => {
             const response = await axios({
                 method: 'get',
-                url: item.link,
+                url: 'item.link',
```
DIYgod: 这是啥？

HenryQW: 估计是不小心按到键盘上了

DIYgod:
Merged, thanks!

PR #2297, opened 2019-06-03
fix: value of rss TTL tag

tuzi3040:
ttl stands for time to live. It's a number of **minutes** that indicates how long a channel can be cached before refreshing from the source.


DIYgod:
Merged, thanks!

PR #2347, opened 2019-06-08
fix: aisixiang ranking

queensferryme:
- 修复路由 `/aisixiang/ranking/:type?/:range?`
- 更新文档，`type` 与排行榜类型对应关系有变

Works on my machine : )


DIYgod:
Merged, thanks!

PR #2494, opened 2019-06-25
docs: change incorrect word order & add translation

hifocus:
- Changed 'Free feel' to 'Feel free' as here appearred to be a word order mistake
- Translate '匿名' into 'Anonymous' to make more sense


DIYgod:
请执行下 npm run format

PR #2577, opened 2019-07-07
fix: remove a redundant if block

HenryQW:
Condition 'false === result' is always false. The value of variable 'result' is originated from the return value of 'parseContent()' defined at line 22


DIYgod:
Merged, thanks!

PR #2578, opened 2019-07-07
fix: remove a redundant condition check

HenryQW:
Condition 'list' is always true. The value of variable 'list' is originated from the array created at line 20 .


DIYgod:
Merged, thanks!

PR #2580, opened 2019-07-07
fix: remove a redundant variable assignment

HenryQW:
Value assigned to variable 'type' at this point is not used afterwards.


DIYgod:
Merged, thanks!

PR #2803, opened 2019-08-09
fix: fix bug introduced in #2796

kidonng:
I screwed up everything 😭 @DIYgod Could you please hurry merge this so it won't affect too many people
I'm really sorry 🙇


kidonng:
👍

PR #2958, opened 2019-08-29
Fix media rss doc

SettingDust:
Fix media rss doc


DIYgod:
a4f8693d889bf73da746a6d5805ef07a202f0363

PR #3021, opened 2019-09-07
Add pubTime for New York Times

QuantumGhost:
The publish time is extracted from the `dateTime` property of `time` element.

### DIFF
--- a/lib/routes/nytimes/morning_post.js
+++ b/lib/routes/nytimes/morning_post.js
@@ -25,13 +25,14 @@ module.exports = async (ctx) => {
     const items = await Promise.all(
         post.map(async (item) => {
             const link = item.link;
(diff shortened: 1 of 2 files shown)


lib/router.js
```diff
@@ -251,7 +251,7 @@ router.get('/yande.re/post/popular_recent/:period', require('./routes/yande.re/p
 
 // 纽约时报
 router.get('/nytimes/morning_post', require('./routes/nytimes/morning_post'));
-router.get('/nytimes/:lang?', require('./routes/nytimes/index'));
+router.get('/nytimes/index/:lang?', require('./routes/nytimes/index'));
```
DIYgod: 看起来没什么意义，反而会让之前的订阅无效

QuantumGhost: OK,我回去把文档改掉

PR #3376, opened 2019-11-02
feat: add naver webtoon

zfanta:
https://comic.naver.com

### DIFF
--- a/lib/routes/webtoons/naver.js
+++ b/lib/routes/webtoons/naver.js
@@ -1,0 +1,39 @@
+const got = require('@/utils/got');
+const cheerio = require('cheerio');
+const domain = 'https://comic.naver.com';
+
+async function getDescription(link) {
+    const { body } = await got.get(link);
+    const $ = cheerio.load(body);
+    return $('#comic_view_area > div.wt_viewer').html();
+}
+
+async function getItems($) {
+    return Promise.all(
+        $('#content > table > tbody > tr')
+            .toArray()
+            .filter((ep) => !ep.attribs.class)
+            .map(async (ep) => ({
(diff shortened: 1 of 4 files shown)


lib/routes/webtoons/naver.js
```diff
@@ -0,0 +1,39 @@
...
+const domain = 'https://comic.naver.com';
+
+async function getDescription(link) {
+    const { body } = await got.get(link);
+    const $ = cheerio.load(body);
+    return $('#comic_view_area > div.wt_viewer').html();
+}
+
+async function getItems($) {
+    return Promise.all(
```
DIYgod: 没有异步操作，不需要使用 Promise.all

zfanta:
I added docs.

lib/routes/webtoons/naver.js
zfanta: At line 16 callback of map has async function. So it needs Promise.all. If Promise.all is removed title and description in all items would be undefined.

DIYgod: Sorry, I miss it.

lib/routes/webtoons/naver.js
```diff
@@ -0,0 +1,39 @@
+const got = require('@/utils/got');
+const cheerio = require('cheerio');
+const domain = 'https://comic.naver.com';
+
+async function getDescription(link) {
+    const { body } = await got.get(link);
```
DIYgod: should use cache for invariable content here, see: https://docs.rsshub.app/en/joinus/#enable-caching

zfanta: I fixed it. please check it.

PR #3581, opened 2019-12-12
feat: Add the Vulture router.

loganrockmore:


lib/router.js
```diff
@@ -2037,4 +2037,16 @@ router.get('/mastodon/timeline/:site/:only_media?', require('./routes/mastodon/t
...
+router.get('/vulture/tv', require('./routes/vulture/tv'));
+router.get('/vulture/movies', require('./routes/vulture/movies'));
+router.get('/vulture/comedy', require('./routes/vulture/comedy'));
+router.get('/vulture/music', require('./routes/vulture/music'));
+router.get('/vulture/tvrecaps', require('./routes/vulture/tvrecaps'));
+router.get('/vulture/books', require('./routes/vulture/books'));
+router.get('/vulture/theater', require('./routes/vulture/theater'));
+router.get('/vulture/art', require('./routes/vulture/art'));
+router.get('/vulture/awards', require('./routes/vulture/awards'));
+router.get('/vulture/video', require('./routes/vulture/video'));
```
DIYgod:
```js
router.get('/vulture/:type', require('./routes/vulture/index'));
```

lib/routes/vulture/art.js
```diff
@@ -0,0 +1,8 @@
+const utils = require('./utils');
+
+module.exports = async (ctx) => {
+    const url = 'https://www.vulture.com/art/';
```
DIYgod:
```js
const url = `https://www.vulture.com/${ctx.params.type}/`;
```

lib/router.js
loganrockmore: Makes total sense, thanks! Fixed!

loganrockmore:
@DIYgod it looks like this failed Travis CI, but for reasons unrelated to the code change.  When I run `npm run test` locally on master, the same test fails.  Any suggestions on what to do here?  Thanks!

DIYgod:
> @DIYgod it looks like this failed Travis CI, but for reasons unrelated to the code change. When I run `npm run test` locally on master, the same test fails. Any suggestions on what to do here? Thanks!

It's not Travis CI failure, it is a git conflict, don't mind.

loganrockmore:
Okay, thank you so much!

PR #3654, opened 2020-01-01
feat: 浙江大学城市学院新闻通知

zhang-wangz:
add rss of zucc_news (浙江大学城市学院新闻通知)
eg: /zucc/news/latest

waiting  add other func...


zhang-wangz:
Why do I just download the code and put in my own crawler structure, so many errors occur ...

zhang-wangz:
sorry for my privious Grammatical errors,and now is ok I think...

PR #3928, opened 2020-02-08
feat: add router yahoo-jp-tv (Yahoo!テレビ)

sakamossan:


lib/routes/yahoo-jp-tv/index.js
```diff
@@ -0,0 +1,27 @@
+const got = require('@/utils/got');
+const cheerio = require('cheerio');
+const querystring = require('querystring');
+
+module.exports = async (ctx) => {
+    const query = querystring.stringify({ q: ctx.params.query });
+    const link = `https://tv.yahoo.co.jp/search/?${query}`;
```
DIYgod: no need querystring, just `https://tv.yahoo.co.jp/search/?q=${ctx.params.query}` is better

PR #4094, opened 2020-02-26
feat: 期刊路由优化，增加 scihub 模式

HenryQW:
1. 增加全局变量 scihub
1. 删除重复冗余代码 elife, pnas @emdoe
1. 修复文档错误
1. 输出 DOI 支持 scihub 模式


yefoenix:
@HenryQW 支持 scihub 可太棒了~
相应的路由要不要加个 badge,类似 bt 和 podcast？

lib/middleware/parameter.js
```diff
@@ -209,6 +209,14 @@ module.exports = async (ctx, next) => {
                 });
                 await Promise.all(tasks);
             }
+
+            // scihub
+            if (ctx.query.scihub) {
+                ctx.state.data.item.map((item) => {
+                    item.link = item.doi ? `https://sci-hub.tw/${item.doi}` : `https://sci-hub.tw/${item.link}`;
```
yech1990: should get the direct link for the pdf file

docs/en/journal.md
```diff
@@ -68,25 +92,31 @@ pageClass: routes
...
+<RouteEn author="yech1990" example="/nature/highlight" path="/journal/nature/highlight" supportScihub="1"/>
 
-### Nature Journal - Research Highlight
+## Proceedings of The National Academy of Sciences (PNAS)
 
-<RouteEn author="yech1990" example="/nature/nature/highlight" path="/nature/nature/highlight" />
+### Latest Articles - Articles by Topic
 
-### Proceedings of The National Academy of Sciences (PNAS) - Latest Articles - ALL
+<RouteEn author="emdoe HenryQW" example="/pnas/Applied Mathematics" path="/journal/pnas/:topic" :paramsDesc="['topic name', 'obtain it from pnas.org (new research in ...). `Latest` will include all.']" supportScihub="1"/>
```
yech1990: Too many preset parameters will increase the requirement for new contributors and the cost of maintenance.

lib/middleware/parameter.js
yech1990: Meanwhile 'sci-hub.tw' has been block by some ISP, which will slow down the rendering of local rss reader.

docs/en/journal.md
HenryQW: On the contrary, keeping 2 copies of the same code snippet decreases the reusability and increases the cost of maintenance.

lib/middleware/parameter.js
HenryQW: > should get the direct link for the pdf file

sci-hub sometimes prompts a captcha, unless that can be addressed, it's not feasible.

HenryQW: > Meanwhile 'sci-hub.tw' has been block by some ISP, which will slow down the rendering of local rss reader.

I will add an ENV for defining the scihub mirror link.

docs/en/journal.md
yech1990: I mean that add `supportScihub` label is not necessary for a simple purpose, which is merely append a prefix to the URL. It should be achieved by passing `:params?` in the router, just in  the same way as other routers. Global parameters and ENV variables should not be invented for special case.

Although a special path `/journal` for the journal section would make the project tidy. But as all the router in this repo are settled by their domain name, those routers for science journal  should not be specialized unless other routes are organized accordingly.

I mean, consistency is more important than convenience in such a huge project.

yech1990:
I mean that creating `supportScihub` label is not necessary for a simple purpose, which is merely append a prefix to the URL. It should be achieved by passing `:params?` in the router, just in the same way as other routers. Global parameters and ENV variables should not be invented for special case.

Although a special path `/journal` for the journal section would make the project tidy. But as all the router in this repo are settled by their domain name, those routers for science journal should not be specialized unless other routes are organized accordingly.

I mean, **consistency is more important than convenience** in such a huge project.

HenryQW:
`supportScihub` is for displaying the badge in the docs, just like BT, podcast and antiCrawler.

I respectfully disagree. I always try to clean the directory and classify relative routes into their own directories. See universities. Think them as namespaces, this doesn’t interfere with the actual route address. While classifying stock routes are already

PR #4199, opened 2020-03-08
add locale module to docs

sgqy:
Can be used, to add a note to titles, whose language/region is different from current UI.

The SVGs is picked from twitter.com

Screenshot_20200308_234413
狗牌（


DIYgod:
I don't think we need it.

PR #4257, opened 2020-03-16
fix: ensure leboncoin `accept` header

### LINKED ISSUE #4244: leboncoin feed doesn't work
### Involved route
https://docs.rsshub.app/en/shopping.html#leboncoin

### What is expected?
rss data for the leboncoin result

the given example  : https://rsshub.app/leboncoin/ad/category=10&locations=Paris_75015 give **Looks like something went wrong
Response code 403 (Forbidden): target website might be blocking our access, you can host your own RSSHub instance for a better usability.**

hoilc:
Fix #4244


Platane:
tested, LGTM

Issue #4537, opened 2020-04-23
[epicgames] Error: this route is empty


junfengP:
### 路由地址
/epicgames/freegames

### 预期是什么？
正常工作

### 实际发生了什么？

Error: this route is empty

### 部署相关信息
rsshub.app

wyzane:
我也遇到了这种情况，详细错误信息如下：
error: Error in /weibo/user/5934019851: Error: this route is empty, please check the original site or <a href="https://github.com/DIYgod/RSSHub/issues/new/choose">create an issue</a>
    at module.exports (/home/service/RSSHub/lib/middleware/parameter.js:19:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async module.exports (/home/service/RSSHub/lib/middleware/template.js:15:5)
    at async /home/service/RSSHub/lib/middleware/api-response-handler.js:143:9
    at async module.exports (/home/service/RSSHub/lib/middleware/api-template.js:2:5)
    at async module.exports (/home/service/RSSHub/lib/middleware/utf8.js:4:5)
    at async module.exports (/home/service/RSSHub/lib/middleware/debug.js:14:5)
    at async module.exports (/home/service/RSSHub/lib/middleware/access-control.js:29:13)
    at async module.exports (/home/service/RSSHub/lib/middleware/header.js:20:5)
    at async module.exports (/home/service/RSSHub/lib/middleware/onerror.js:20:9)

PR #4550, opened 2020-04-24
fix: route epicgames

junfengP:
1. change api
2. fix a bug (打折游戏会被列入免费游戏，并且title === undefined，将之移除)


junfengP:
close https://github.com/DIYgod/RSSHub/issues/4537

PR #5340, opened 2020-08-03
feat: add CuriousCat

lucasew:
After some time using the public instance in my personal bots and knowing how awesome this app is I missed support to CuriousCat, after some inspecting (less than 1 hour precisely) I have found a API that the frontend requests against to get its data then I found some opportunity to contribute with you.

I am using NixOS and I tough it will be easier to setup a nix-shell environment than setup a full dev container environment using VS Code so I linked that nix-shell script if anyone else wants to use it to contribute. It's as easier as `cd` into the cloned repo folder and running `nix-shell` so nodejs and yarn will be in your path, then from this you can run `code .` to start vscode from this environment.


HenryQW:
Thanks for the PR. I would suggest that the `shell.nix` goes into the doc instead, as it's not a common use case.

PR #5389, opened 2020-08-10
Fix? #5385. Force allow `solidot` cert, while `WoTrust` is unclear.

sgqy:
It may be a dangerous choice.
Should we only use this locally?


HenryQW:
Local instances can set env "NODE_TLS_REJECT_UNAUTHORIZED=0" to bypass it or modify the code like in this PR. I will not merge unsafe codes into the demo.

PR #5555, opened 2020-08-28
fix: remove pornhub preview video

I2IMk:

Close #

/pornhub/category_url/
/pornhub/category_url/popularwithwomen

### DIFF
--- a/lib/routes/pornhub/category_url.js
+++ b/lib/routes/pornhub/category_url.js
@@ -10,7 +10,7 @@ module.exports = async (ctx) => {
     const list = $('.nf-videos .videoBox');
 
     ctx.state.data = {
-        title: $('title').text(),
+        title: $('title').first().text(),
         link: link,
         item:
             list &&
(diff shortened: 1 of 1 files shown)


HenryQW:
Any reason why?

I2IMk:
@HenryQW 
These preview videos extracted will expire in (exactly) two hours, which is too short for feed. 
Some video urls have indicated this: `https://ew.phncdn.com/videos/../360P_360K_.._fb.mp4?validfrom=1598770991&validto=1598778191&rate=150k&burst=1000k&hash=..`

HenryQW:
Is it possible to obtain a permanent src?

I2IMk:
I don't think so. Since these preview links look so similar to streaming ones, the restrictions imposed should be similar too. 

The [webmaster API](https://github.com/pionxzh/Pornhub.js/blob/master/src/webmaster/index.js) won't give a preview video as well, e.g. `https://www.pornhub.com/webmasters/video_by_id?id=ph5ecf18b141646&thumbsize=large_hd`

I have replaced preview videos to static thumbnails.

PR #5600, opened 2020-09-04
feat: add jian-ning blog

changlan:

Close #

/blogs/jianning

### DIFF
--- a/assets/radar-rules.js
+++ b/assets/radar-rules.js
@@ -2262,7 +2262,7 @@
             },
         ],
     },
-    'wenxuecity.com/': {
+    'wenxuecity.com': {
         _name: '文学城',
         blog: [
             {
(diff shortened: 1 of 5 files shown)


assets/radar-rules.js
```diff
@@ -2251,14 +2251,14 @@
             },
         ],
     },
-    'manxiaosi.com': {
```
HenryQW, line 2254: 为什么覆盖了别的路由？

HenryQW (changes requested):

assets/radar-rules.js
changlan: Fixed

changlan:
@HenryQW Fixed. I don't quite get the error message from CodeFactor -- does it basically mean there are too many entries in the `radar-rules.js` file?

PR #5747, opened 2020-09-28
fix: (semi-fix) login to NGA to prevent 403 error.

sgqy:

Cannot open bbs-thread to read. Rejected by a redirect page with response header HTTP 403.

`http://127.0.0.1:1200/nga/post/22708129`


HenryQW:
为什么不直接把相关配置加入`lib/config.js`?

PR #5939, opened 2020-10-19
feat: add mit ocw route

dwemerx:

Doesn't seem to be any

/mit/ocw-top

## Comments

I'd like to add a couple more routes and have a few questions. I'd be happy if someone helps me figure this out:

1. Is there anything I should change so it'll be easier to maintain them?
2. I noticed there are a bunch of university routes that aren't listed in the docs. Does it mean that all broken routes get removed until someone fix them?
3. Once I got the feed working, it started to generate a field called "guid" with "isPermaLink="false"". Is it something I should touch or just let it be?

Thanks!

### DIFF
--- a/docs/en/university.md
+++ b/docs/en/university.md
@@ -6,6 +6,10 @@ pageClass: routes
 
 ## MIT
 
+### MIT OCW Most popular courses of the month
+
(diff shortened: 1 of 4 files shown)


HenryQW:
1. Don't think so.
1. Not every route has an EN-version doc, if that's what you mean.
1. That's the default RSS specification, nothing to worry about.

Issue #6032, opened 2020-10-28
小米众筹、infoQ、腾讯柠檬的rss貌似失效了


csintest:
- 完整路由地址，包含所有必选与可选参数
/mi/crowdfunding
/infoq/recommend
/Tencent/lemon

- 预期是什么
能订阅更新。

- 实际发生了什么
RSS app上发现订阅未更新，订阅的标题也变为了rsshub服务器域名。
发现后台日志报错。

- 部署相关信息

| Env                | Value         |
| ------------------ | ------------- |
| OS                 |        CentOS  7.7.1908     |
| Node version       |        v12.18.3       |
| if Docker, version |       -        |

- 额外信息（日志、报错等）

error: Error in /mi/crowdfunding: Error: this route is empty, please check the original site or <a href="https:/
/github.com/DIYgod/RSSHub/issues/new/choose">create an issue</a>
    at module.exports (/home/RSSHub/lib/middleware/parameter.js:20:19)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async module.exports (/home/RSSHub/lib/middleware/anti-hotlink.js:33:5)
    at async module.exports (/home/RSSHub/lib/middleware/template.js:15:5)
    at async /home/RSSHub/lib/middleware/api-response-handler.js:143:9
    at async module.exports (/home/RSSHub/lib/middleware/api-template.js:2:5)
    at async module.exports (/home/RSSHub/lib/middleware/utf8.js:4:5)
    at async module.exports (/home/RSSHub/lib/middleware/header.js:20:5)
    at async module.exports (/home/RSSHub/lib/middleware/debug.js:14:5)
    at async grant (/home/RSSHub/lib/middleware/access-control.js:23:13)
    at async module.exports (/home/RSSHub/lib/middleware/access-control.js:31:20)
    at async module.exports (/home/RSSHub/lib/middleware/onerror.js:20:9)

error: Error in /infoq/recommend: TypeError: Cannot read property 'map' of undefined
    at parseToSimpleTexts (/home/RSSHub/lib/routes/infoq/utils.js:45:13)
    at parseToSimpleText (/home/RSSHub/lib/routes/infoq/utils.js:43:40)
    at Object.link (/home/RSSHub/lib/routes/infoq/utils.js:76:30)
    at /home/RSSHub/lib/routes/infoq/utils.js:82:36
    at Array.map (<anonymous>)
[...]

PR #6033, opened 2020-10-29
fix: 腾讯柠檬精选 Mac Apps

nczitzk:
## 完整路由地址 / Example for the proposed route(s)

/tencent/lemon

PR #6034, opened 2020-10-29
fix: 腾讯柠檬 rss 失效 #6032

laampui:

#6032

/Tencent/lemon


laampui:
close due to duplicate #6033

PR #6043, opened 2020-10-29
docs: replace aggregator with generator

johan456789:
RSS readers like Feedly and Inoreader are [aggregators](https://en.wikipedia.org/wiki/Comparison_of_feed_aggregators?oldformat=true). RSSHub is a feed generator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 ## Introduction
 
-RSSHub is an open source, easy to use, and extensible RSS feed aggregator, it's capable of generating RSS feeds from pretty much everything.
+RSSHub is an open source, easy to use, and extensible RSS feed generator. It's capable of generating RSS feeds from pretty much everything.
 
 RSSHub delivers millions of contents aggregated from all kinds of sources, our vibrant open source community is ensuring the deliver of RSSHub's new routes, new features and bug fixes.
 


HenryQW:
It's a hub, so I guess it's fair to say it aggregates? Also RSSHub has more to offer, there are options to achieve [filtering and others](https://docs.rsshub.app/en/parameter.html)

johan456789:
An [aggregator](https://en.wikipedia.org/wiki/Aggregator?oldformat=true) put things from many sources into one place. There are no other sources for the RSS feeds in RSSHub. They are defined/generated by the author. RSSHub is the source.

> RSSHub 是一个开源、简单易用、易于扩展的 RSS **生成器**

And "generator" literally matches the description in Chinese.

I think it would be best to use generator. Using aggregator is confusing, as it means a totally different thing already.

From Wikipedia [News aggregator](https://en.wikipedia.org/wiki/News_aggregator?oldformat=true) page:

> In computing, a news aggregator, also termed a **feed aggregator**, feed reader, news reader, RSS reader or simply an **aggregator**, is client software or a web application that aggregates syndicated web content such as online newspapers, blogs, podcasts, and video blogs (vlogs) in one location for easy viewing.

HenryQW:
I'd rather focus on other issues than being pedantic...

PR #6057, opened 2020-10-30
feat: add wsj opinion

oppilate:
`/wsj/en-us/opinion`


oppilate:
Sorry that I didn't change the route tag in en doc to `RouteEn`. Will send a pr later.

BTW it's not jus wsj opinion. Maybe I didn't make it clear... it supports all categories as I said in the doc.

wj-c:
What's the route for channel like market news? I don't seem to get it right. There is no apparent relation of routes to office rss.

PR #6353, opened 2020-12-05
fix: Instagram carousel exception

oppilate:
## 该 PR 相关 Issue / Involved 
- Not compatible with vercel.app. Not possible to login (because of variable's lifecycle?)
- Carousel caused exception

`/instagram/user/ladygaga`


oppilate:
In the latest commit, hopefully it's possible to save the login status for non-vercel environments and login at every request on vercel.app

oppilate:
Sorry, don't know how to fix that DeepScan complaint

oppilate:
"Solved" by disabling that complaint.

HenryQW:
Actually, the compatibility with vercel is not required, you don't have to preserve the login status.

proletarius101:
Well, but I use it there... Because it's faster and free.

oppilate:
Well, but I use it there :rofl:  It's faster and free. And why not?

oppilate:
And interestingly, the build fails. But I didn't touch anything in `./docs/game.md`?

HenryQW:
> And interestingly, the build fails. But I didn't touch anything in `./docs/game.md`?

That was something else in the master which was fixed.



> Well, but I use it there 🤣  It's faster and free. And why not?

It's tricky to test that out, I will have to revisit this another day.

oppilate:
> you don't have to preserve the login status.

And, to prevent misunderstanding, Vercel is more request-based which means it needs to login per request. On docker or other type of deplyments, it only needs one login per app session.

Issue #6413, opened 2020-12-10
Instagram route returning "Error message: Unknown error"


igor-akhmetov:
- The involved route, with all required and optional parameters

Instagram route via my instance with correct Instagram credentials, e.g. myrsshub/instagram/user/airbus

- What is expected

The route return a feed.

- What is actually happening

I get an error:
Route requested: /instagram/user/airbus
Error message: Unknown error

- Self-deployment information

Latest docker image on my VPS host.

- Additional info (logs errors etc)

The log just contains "Unknown error" lines. It seems like after RSSHub is started the first Instagram route I request goes through and returns a feed, but after that the other Instagram route requests return the error.

HenryQW:
The problem is the challenge prompted by Instagram when you log in on a new device. You will have to log in via Instagram app, click "this was me". If you do not do this, your subsequent login will likely be challenged by a security code, unfortunately it is very tricky to submit the code via your RSSHub. 

After clicking "this was me" 3 times, Instagram seems me remember my RSSHub and no longer causes the error.

igor-akhmetov:
Thanks! That was my guess, I log on from my PC but got no confirmation prompt on the website. I guess I will try with the app.

stjohnjohnson:
I'm still having the same issue.  I can load instagram for 2-3 users and then I get "unknown error."

Issue #6268, opened 2020-11-26
Instagram RSS feed for #haschtags


lopesfleonardo:
(https://docs.rsshub.app) [issue](https://github.com/DIYgod/RSSHub/issues)
Is there a way to use RSSHub for instagram hashtags? 
I tried, https://rsshub.app/instagram/tag/fashion and https://rsshub.app/instagram/tag/restaurantedogringo/date but they failled

If not, could you improve that, please?? It is possible to show first the the recent posts?

PR #6434, opened 2020-12-13
fix: try to fix instagram account suspension

oppilate:

Sorry to spam on Instagram again... It is reported that the bridge could lead to account suspension. And I realize I didn't add the post login flow
````javascript
            process.nextTick(async () => await ig.simulate.postLoginFlow());
````
which might lead to the case. So here is the patch. More feedbacks are welcomed.

````
/instagram/user/ladygaga
````


HenryQW:
I look into the private api library, it looks like it supports challenge (code via SMS or email) to avoid challenge errors #6413, but I'm not sure if that can be implemented easily. Maybe that's worth investigating?

HenryQW:
And maybe help implement features like hashtag #6268

oppilate:
For #6413, maybe show a challenge page at `/instagram/challenge` ? Not a trivial job.

PR #6445, opened 2020-12-14
Scmp series

oppilate:

Depends on  #6444

`/scmp/318198`

There's an item with link containing `series`. It would break the feed. This patch fixes the problem.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,4 +1,6 @@
 require('dotenv').config();
+const UserAgent = require('user-agents');
+
 let envs = process.env;
 let value;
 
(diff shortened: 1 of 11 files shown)


HenryQW:
This PR should not include the UA changes.

oppilate:
Unfortuatenly there's a change of the UA variable. So it has to include the UA change or if the UA change is rejected it has to rollback that part.

NeverBehave:
Hi, I believe we may have to separate these two: as the UA changes won't come in the short time. 

If you still want to land the rss changes, it may have to done in the "ugly" way for now. Please advise.

oppilate:
Sure. Since the ua issue stucks, this issue should not depend on that anymore. Will do the job tomorrow.

NeverBehave:
> Sure. Since the ua issue stucks, this issue should not depend on that anymore. Will do the job tomorrow.

Cool, I will close this pr for now. open a new one when finished.

PR #6514, opened 2020-12-24
feat: added DailyArt & fix: error occurred when without user-agent

zphw:

route: /dailyart/:language?
example: /dailyart/zh

- language: optional, indicating the text language. Languages supported include en, es, fr, de, it, zh, jp, etc. English by default.


zphw:
Also fixed an error occurred when requesting without user-agent that can cause `TypeError: Cannot read property 'includes' of undefined`

PR #6544, opened 2020-12-27
fix: disable default debug info

zphw:
Changed default debugInfo config to false to avoid potential privacy leakage.

RSSHub displays client IP within debug information when debugInfo is set to true, and since all pull request will be automatically deployed with Vercel, anyone, including contributors, who uses the deployment for any purpose would passively disclose their IP addresses to the instance in public. As per the principle of least privilege, therefore, users should not be granted the access to the information by default. Instead, this configuration should be manually changed for debugging purposes where needed.


HenryQW:
Good point.

PR #6558, opened 2020-12-28
feat: add Uestc sice notice

huyyi:

Close #

https://rsshub.app/uestc/sice


zphw:
Oops. You can use `let date;` to initialize the variable before using for CodeFactor issues.

PR #6588, opened 2021-01-03
feat: add fulltext to the article route of caixin

lonelykid:

None

/caixin/article

Although most of the articles are not free, the free part of each article is still informative. I'd like to have it in the RSS. If people don't like it, I can add a switch and return without the text by default.

### DIFF
--- a/lib/routes/caixin/article.js
+++ b/lib/routes/caixin/article.js
@@ -1,4 +1,5 @@
 const got = require('@/utils/got');
+const cheerio = require('cheerio');
 
 module.exports = async (ctx) => {
     const response = await got({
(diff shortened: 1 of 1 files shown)


lonelykid:
@NeverBehave Just find out this has conflicts with the ``caixin/latest`` route. Sorry about that. He cached the final item object. This caches only the main content. That latest api is always faster(but fewer contents). so [object Object] is displayed in the route. Should I add a tag to the key to make it different or change one of them to cache the same thing.

NeverBehave:
> @NeverBehave Just find out this has conflicts with the `caixin/latest` route. Sorry about that. He cached the final item object. This caches only the main content. That latest api is always faster(but fewer contents). so [object Object] is displayed in the route when requesting the same url. Should I add a tag to the key to make it different or change one of them to cache the same thing.

I have revert this pr for now.

This is an interesting problem tho, I'd say use a different key for now. We need to standardize a little bit about our scripts.

PR #6595, opened 2021-01-03
feat: add homepage route for Technology Review

zphw:

/technologyreview


NeverBehave (changes requested):
Could add this route to CN documentation as well?

zphw:
there’s already an entry called “MIT 科技评论” that seems the chinese version of this so i supposed no need to do so

NeverBehave (approved):
LGTM

PR #6602, opened 2021-01-04
Fix icity item link

yueyericardo:
Otherwise guid will be the same for all entries

/icity/sai


NeverBehave (approved):
LGTM

side note: Maybe url resolve should be on manual

PR #6741, opened 2021-01-20
fix(config): debug not shown when DEBUG_INFO set true

tuzi3040:

Close #

```routes
NOROUTE
```

- [ ] 这个PR中包含了新的路由吗? Does this PR add new route?
  - 如果有, 请完成检查列表. If yes, please finish the check list
  - **如果你的PR符合下方某个事项, 也请注明. If any of the checklist item meets your PR, please fill it out.**
  - [x] <- 这样打勾
- [ ] 是否提供了文档? Documentation provided?
  - [ ] 是否提供了英文文档? EN Documentation provided?
- [ ] 是否支持全文获取? Is this RSS Script support fulltext?
  - [ ] 如果全文获取中需要访问文章链接, 是否使用了缓存? If fulltext requires to fetch detail pages, is cache used in the process?
  - [缓存说明](https://docs.rsshub.app/joinus/#ti-jiao-xin-de-rsshub-gui-ze-bian-xie-jiao-ben-shi-yong-huan-cun) | [How to use cache](https://docs.rsshub.app/joinus/#ti-jiao-xin-de-rsshub-gui-ze-bian-xie-jiao-ben-shi-yong-huan-cun)
- [ ] 目标是否有明显的反爬/频率限制? Is there any sign of anti-bot or rate limit?
  - [ ] 如果有, 是否有对应的措施? (延长缓存时间, 写文档说明, etc.) If yes, do your code reflect this sign? (e.g. write documentations, use long cache time)
- [ ] 是否引入的新的包? Any new package introduced?
  - 如果有, 请说明原因. If yes, please state your reason
- [ ] 是否使用了`Puppeteer`? Make use of `Puppeteer`?
  - 如果有, 请说明原因. If yes, please state your reason
  

## 说明 / Note

Debug info is not shown when DEBUG_INFO is set to 'true'.

### DIFF
--- a/docs/en/install/README.md
+++ b/docs/en/install/README.md
@@ -423,7 +423,7 @@ See the relation between access key/code and white/blacklisting.
 
 `REQUEST_RETRY`: retries allowed for failed requests, default to `2`
 
-`DEBUG_INFO`: display route information on homepage for debugging purpose, default to `false`
+`DEBUG_INFO`: display route information on homepage for debugging purpose. When set to neither `true` nor `false`, use parameter `debug` to enable display, eg: <https://rsshub.app/?debug=value_of_DEBUG_INFO> . Default to `false`
 
 `NODE_ENV`: display error message on pages for authentication failing, default to `production` (i.e. no display)
 
--- a/docs/install/README.md
+++ b/docs/install/README.md
@@ -451,7 +451,7 @@ RSSHub 支持使用访问密钥 / 码，白名单和黑名单三种方式进行�
 
 `REQUEST_RETRY`: 请求失败重试次数，默认 `2`
 
-`DEBUG_INFO`: 是否在首页显示路由信息，默认 `false`
+`DEBUG_INFO`: 是否在首页显示路由信息。值为非 `true` `false` 时，在请求中带上参数 `debug` 开启显示，例如：<https://rsshub.app/?debug=value_of_DEBUG_INFO> 。默认 `false`
 
 `NODE_ENV`: 是否显示错误输出，默认 `production` （即关闭输出）
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -39,8 +39,8 @@ const calculateValue = () => {
         ua: envs.UA || 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.181 Safari/537.36',
         listenInaddrAny: envs.LISTEN_INADDR_ANY || 1, // 是否允许公网连接，取值 0 1
         requestRetry: parseInt(envs.REQUEST_RETRY) || 2, // 请求失败重试次数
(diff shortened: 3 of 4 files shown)


NeverBehave (changes requested):
Could you also change the default value in `config.js`? it is using false instead of 'false'

tuzi3040:
> Could you also change the default value in `config.js`? it is using false instead of 'false'

You do understand changing boolen `false` to string `'false'` has no effect on the final result, right?
And you request this change just for code style?
I just want to make sure we are on the same page.

NeverBehave:
> > Could you also change the default value in `config.js`? it is using false instead of 'false'
> 
> You do understand changing boolen `false` to string `'false'` has no effect on the final result, right?
> And you request this change just for code style?
> I just want to make sure we are on the same page.

Even though js is weak type language here, this option is defined as a string and we keep its type the same here. 
It is also a tip when people reading the config file knowing it is a string option rather than a boolean.

tuzi3040:
> > > Could you also change the default value in `config.js`? it is using false instead of 'false'
> > 
> > 
> > You do understand changing boolen `false` to string `'false'` has no effect on the final result, right?
> > And you request this change just for code style?
> > I just want to make sure we are on the same page.
> 
> Even though js is weakly typed language here, this option is defined as a string and we keep its type the same here.
> It is also a tip when people reading the config file knowing it is a string option rather than a boolean.
> tbh, I am confused at first.

Done.

DIYgod:
a0dd6756dad5b682e88863400c594d402a26e19f 6506c39239e5bd35a27889d92d46acca9a048c63 I have reset the default value of debugInfo to true, and remove `hot ip`, so maybe this pr is no longer needed

NeverBehave:
If the inconsistency doesn't matter (true and 'false'), this pr could be closed.

tuzi3040:
> a0dd6756dad5b682e88863400c594d402a26e19f 6506c39239e5bd35a27889d92d46acca9a048c63 I have reset the default value of debugInfo to true, and remove `hot ip`, so maybe this pr is no longer needed

Though the default value has been reset to `true` now, when DEBUG_INFO set to string `'true'`, route info is not always displayed. Which requires `?debug=true` instead, may be confusable. But it seems intentional in comment.
And what about use NODE_ENV to control hot ip display, like detailed error output requires NODE_ENV not `'production'`.

DIYgod:
> Though the default value has been reset to `true` now, when DEBUG_INFO set to string `'true'`, route info is not always displayed. Which requires `?debug=true` instead, may be confusable. But it seems intentional according to comment?...
> And what about use NODE_ENV to control hot ip display, like detailed error output requires NODE_ENV not `'production'`.

You are right, I merged your changes 1a463e09c99c1c70b66adad43b950a9241d09c09

As for controlling the IP display, I think there are privacy issues, it would be better to remove it completely, the owner should view it through the backend logs instead.

tuzi3040:
> You are right, I merged your changes [1a463e0](https://github.com/DIYgod/RSSHub/commit/1a463e09c99c1c70b66adad43b950a9241d09c09)

thx
> As for controlling the IP display, I think there are privacy issues, it would be better to remove it completely, the owner should view it through the backend logs instead.

That's true. I agree with you.
I'll close this pr now.

PR #8118, opened 2021-08-31
feat(route): add eprice

TonyRL:

Close #

```routes
/eprice
/eprice/hk
/eprice/tw
```

- [x] New Route
- [x] Documentation
  - [x] CN
  - [ ] EN
- [x] 全文获取 fulltext
  - [x] Use Cache
- [x] 日期和时间 date and time
  - [x] 可以解析 Parsed
  - [x] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
Regarding pubDate, upstream provides them in the format of `Tue, 31 Aug 2021 18:29:04 +0800`. No extra timezone adjustment is needed.

### DIFF
--- a/docs/new-media.md
+++ b/docs/new-media.md
@@ -224,6 +224,18 @@ pageClass: routes
 
 </Route>
 
+## ePrice
+
+<Route author="TonyRL" example="/eprice/tw" path="/eprice/:region?" :paramsDesc="['地区，预设为tw']">
+
+地区：
+
+| hk   | tw   |
+| ---- | ---- |
+| 香港 | 台湾 |
+
+</Route>
+
 ## Esquirehk
 
 ### Tag
--- a/docs/new-media.md
+++ b/docs/new-media.md
@@ -226,7 +226,7 @@ pageClass: routes
 
 ## ePrice
 
-<Route author="TonyRL" example="/eprice/tw" path="/eprice/:region?" :paramsDesc="['地区，预设为tw']">
+<Route author="TonyRL" example="/eprice/tw" path="/eprice/:region?" :paramsDesc="['地区，预设为 tw']">
 
 地区：
 
--- a/lib/routes/eprice/rss.js
+++ b/lib/routes/eprice/rss.js
@@ -61,7 +61,7 @@ module.exports = async (ctx) => {
                 delete item.enclosure;
                 delete item.isoDate;
 
-                item.description = $('div.user-comment-block').html() || $('div.content').html();
+                item.description = $('div.user-comment-block').html() ?? $('div.content').html();
                 item.pubDate = parseDate(item.pubDate);
 
                 return item;
--- a/lib/routes/eprice/rss.js
+++ b/lib/routes/eprice/rss.js
@@ -9,55 +9,54 @@ module.exports = async (ctx) => {
     const feed = await parser.parseURL(`https://www.eprice.com.${region}/news/rss.xml`);
 
     const items = await Promise.all(
-        feed &&
-            feed.items.map(async (item) => {
-                item.link = item.link.replace(/^http:\/\//i, 'https://');
-                const cache = await ctx.cache.get(item.link);
-                if (cache) {
-                    return JSON.parse(cache);
+        feed.items.map(async (item) => {
+            item.link = item.link.replace(/^http:\/\//i, 'https://');
+            const cache = await ctx.cache.get(item.link);
+            if (cache) {
+                return JSON.parse(cache);
+            }
+
+            const response = await got({
+                method: 'get',
+                url: item.link,
+                headers: {
+                    Referer: `https://www.eprice.com.${region}`,
+                },
+            });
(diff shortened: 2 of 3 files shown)


docs/new-media.md
```diff
@@ -224,6 +224,18 @@ pageClass: routes
 
 </Route>
 
+## ePrice
+
+<Route author="TonyRL" example="/eprice/tw" path="/eprice/:region?" :paramsDesc="['地区，预设为tw']">
```
SettingDust, line 229:
```suggestion
<Route author="TonyRL" example="/eprice/tw" path="/eprice/:region?" :paramsDesc="['地区，预设为 tw']">
```

lib/routes/eprice/rss.js
```diff
@@ -0,0 +1,70 @@
+const got = require('@/utils/got');
+const parser = require('@/utils/rss-parser');
+const cheerio = require('cheerio');
+const { parseDate } = require('@/utils/parse-date');
+
+module.exports = async (ctx) => {
+    const region = ctx.params.region || 'tw';
```
SettingDust, line 7:
```suggestion
    const region = ctx.params.region ?? 'tw';
```

lib/routes/eprice/rss.js
```diff
@@ -0,0 +1,70 @@
...
+
+    const feed = await parser.parseURL(`https://www.eprice.com.${region}/news/rss.xml`);
+
+    const items = await Promise.all(
+        feed.items.map(async (item) => {
+            item.link = item.link.replace(/^http:\/\//i, 'https://');
+            const cache = await ctx.cache.get(item.link);
```
SettingDust, line 14: tryGet

lib/routes/eprice/rss.js
```diff
@@ -0,0 +1,70 @@
...
+            $('img').each((_, e) => {
+                if ($(e).attr('data-original')) {
+                    $(e).after(`<img src="${$(e).attr('data-original')}" alt="${$(e).attr('alt')}" title="${$(e).attr('title')}">`);
+                    $(e).remove();
+                }
+            });
+
+            const single = {
+                title: item.title,
+                description: $('div.user-comment-block').html() || $('div.content').html(),
```
SettingDust, line 50:
```suggestion
                description: $('div.user-comment-block').html() ?? $('div.content').html(),
```

lib/routes/eprice/rss.js
```diff
@@ -0,0 +1,70 @@
...
+            $('div.text-left').remove();
+            $('div.signature').remove();
+
+            // fix lazyload image
+            $('img').each((_, e) => {
+                if ($(e).attr('data-original')) {
+                    $(e).after(`<img src="${$(e).attr('data-original')}" alt="${$(e).attr('alt')}" title="${$(e).attr('title')}">`);
+                    $(e).remove();
+                }
+            });
```
SettingDust, line 46: 替换原有属性即可

SettingDust (changes requested):

lib/routes/eprice/rss.js
TonyRL: Resolved in https://github.com/DIYgod/RSSHub/pull/8118/commits/7f3ab54d3c029296cc0bacef171f5357533dc2da

lib/routes/eprice/rss.js
TonyRL: Resolved in https://github.com/DIYgod/RSSHub/pull/8118/commits/551e57edc0f4d9bbec8632fe91a9297e7ec396dc

lib/routes/eprice/rss.js
```diff
@@ -0,0 +1,80 @@
...
+                    e = $(e);
+                    if (e.attr('data-original')) {
+                        e.attr('src', e.attr('data-original'));
+                        e.removeAttr('class');
+                        e.removeAttr('data-original');
+                        e.removeAttr('data-title');
+                        e.removeAttr('data-thumbnail');
+                        e.removeAttr('data-url');
+                    }
+                });
```
SettingDust, line 54: 判断属性 filter 更好些，移除属性用白名单遍历移除更好些

SettingDust (approved):

lib/routes/eprice/rss.js
TonyRL: Your suggested change will have something like
```javascript
$('img').filter(... => ...condition...). // filter loop: O(n)
each((_, e) => { // each loop: O(n)
// attr swap: O(1)
//
// loop through attrs and whitelisting like https://stackoverflow.com/a/3643317
// O(n)
});
```

By considering the time complexity(You should know that there is a really strict function timeout on Vercel and other cloud platforms), the present solution(only one each loop: O(n) ) is better than nested loops (time complexity > O(n) ). 

Someone from [stackoverflow](https://stackoverflow.com/a/5929629) also did profiling and shows that less loop is faster.

PR #8521, opened 2021-11-12
fix(route): 东方网-上海新闻

saury:

```routes
/eastday/sh
```


saury:
Got conflict. Move to #8523

PR #8660, opened 2021-12-04
up biquge url

weqopy:

```routes
/novel/biquge/152_152598
```

## 说明 / Note

update url of the site.

### DIFF
--- a/lib/routes/novel/biquge.js
+++ b/lib/routes/novel/biquge.js
@@ -3,7 +3,7 @@ const cheerio = require('cheerio');
 const iconv = require('iconv-lite');
 const asyncPool = require('tiny-async-pool');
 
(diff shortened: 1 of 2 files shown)


lib/routes/novel/biquge.js
```diff
@@ -3,7 +3,7 @@ const cheerio = require('cheerio');
 const iconv = require('iconv-lite');
 const asyncPool = require('tiny-async-pool');
 
-const baseUrl = 'https://www.biquge5200.com/';
+const baseUrl = 'https://www.biquge5200.net/';
```
TonyRL, line 6:
```suggestion
const baseUrl = 'http://www.biquge5200.net/';
```
The site does not have HTTPS enabled.

TonyRL (changes requested):

PR #8700, opened 2021-12-08
fix: remove entity decoding for item html description

### LINKED ISSUE #8699: HTML entity decoding problem of item description
### Routes

All

### Full routes

All

### Related documentation

No

### What is expected?

A message with text `test message: <p>test</p> test` should be converted into item with description `<p>test message: &lt;p&gt;test&lt;/p&gt; test</p>` because item description is HTML and will be later parsed by other HTML parsers, which requires it to be correctly escaped.

A message with text `test message: <p>test</p> test` will actually be converted into item with description `<p>test message: <p>test</p> test</p>`, in which the app processed already-escaped HTML crawled from original websites with entity decoding and resulted in item HTML description that seems not to be escaped.

An example can be found with Telegram post https://t.me/myl7s/722 with route /telegram/channel/myl7s . However, it affects all routes with item description.

### Deployment information

Self-hosted

_No response_

### Additional info

```shell
A PR has been provided simultaneously to provide (or propose) a fix to the issue.
```

- [X] I have searched [existing issues](https://github.com/DIYgod/RSSHub/issues) to ensure this bug has not already been reported

myl7:

Close #8699

```routes
NOROUTE
```

- [ ] New Route
- [ ] Documentation
  - [ ] CN
  - [ ] EN
- [ ] 全文获取 fulltext
  - [ ] Use Cache
- [ ] 日期和时间 date and time
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note

Updated parameter cleaning of `item.description` in `lib/middleware/parameter.js`

### DIFF
--- a/lib/middleware/parameter.js
+++ b/lib/middleware/parameter.js
@@ -112,7 +112,7 @@ module.exports = async (ctx, next) => {
                         $ele.removeAttr(e);
                     });
                 });
-                item.description = entities.decodeXML($('body').html() + '') + (config.suffix || '');
+                item.description = $('body').html() + '' + (config.suffix || '');
             }
             return item;
         };


SukkaW (commented):
Will it cause breaking changes?

myl7:
I have tried to `git blame` it. The entity encoding is added at 90a87b172 which is a refactor other than fix/feat. Further reason tracing may need to ask the commit author @kt286 .

One can be ensured to be affected according to the doc is [parameter `filter_description` and `filterout_description`](https://docs.rsshub.app/parameter.html#nei-rong-guo-lu) handling. Comparing to setting `filter_description=<p>test</p>` previously which is the literal value, now users need to set `filter_description=&lt;p&gt;test&lt;/p&gt;` to do the same filter.

When router developers need to add `<` into `item.description`, they are (obviously) more likely to add `<` straightforward other than `&lt;` (the two will result in the same thing in the item description output). So there shouldnt be problems in router code.

As for other projects using the app and relying on the bug, it means parsing (even rendering) publisher messages without escaping, which is dangerous (when rendering is included, there may be XSS problems). The number of these unwise projects should be very small and it is fair to introduce the breaking change to them.

PR #8775, opened 2021-12-25
点击bilibili视频描述，打开app观看

smilingpoplar:

```routes
/bilibili/user/video
```

## 说明 / Note

因为ios上embed播放器没有倍速播放，所以加个跳转，打开app观看


DIYgod (changes requested):
It is better to add a  `<a src="xxx">APP 打开</a>` than to make the whole description a link

smilingpoplar:
> It is better to add a `<a src="xxx">APP 打开</a>` than to make the whole description a link

我个人使用是因为不想特意去找"APP打开"几个字，整个description的点击范围更大，跳转更方便

PR #8921, opened 2022-01-24
fix(route)(dxy);add(route)(sh-wsjkw)

zcf0508:

```routes
/coronavirus/dxy 
/sh-wsjkw
```

## 说明 / Note

修复丁香园数据获取失败的问题
添加上海卫健委疫情通报


TonyRL (changes requested):
https://github.com/DIYgod/RSSHub/blob/ef3086f682afb6464402f6a3d946bb961e962f51/lib/v2/sh-wsjkw/index.js#L5
```suggestion
    const url = `https://wsjkw.sh.gov.cn/yqtb/index.html`;
```
HTTPS is preferred

TonyRL (commented):
https://github.com/DIYgod/RSSHub/blob/ef3086f682afb6464402f6a3d946bb961e962f51/docs/other.md#L926

Since your route has gov.cn at the end, consider putting it in `/v2/gov` folder and changing to the path to `/gov/shanghai/wsjkw/yqtb` like other gov-related routes.

PR #8971, opened 2022-01-31
ci: release docker in source only

liyishuai:

Close #

```routes
NOROUTE
```

- [ ] New Route
- [ ] Documentation
  - [ ] CN
  - [ ] EN
- [ ] 全文获取 fulltext
  - [ ] Use Cache
- [ ] 日期和时间 date and time
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -14,6 +14,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'DIYgod'
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -25,7 +26,6 @@ jobs:
         uses: docker/setup-buildx-action@v1
 
       - name: Login to DockerHub
-        if: github.repository_owner == 'DIYgod'
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
@@ -33,7 +33,6 @@ jobs:
 
       - name: Extract Docker metadata
         id: meta
-        if: github.repository_owner == 'DIYgod'
         uses: docker/metadata-action@v3
         with:
           images: ${{ secrets.DOCKER_USERNAME }}/rsshub
@@ -43,7 +42,6 @@ jobs:
           flavor: latest=false
 
       - name: Build and push Docker image
-        if: github.repository_owner == 'DIYgod'
         uses: docker/build-push-action@v2
         with:
           context: .
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -25,6 +25,7 @@ jobs:
         uses: docker/setup-buildx-action@v1
 
       - name: Login to DockerHub
+        if: github.repository_owner == 'DIYgod'
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
@@ -32,6 +33,7 @@ jobs:
 
       - name: Extract Docker metadata
         id: meta
+        if: github.repository_owner == 'DIYgod'
         uses: docker/metadata-action@v3
         with:
           images: ${{ secrets.DOCKER_USERNAME }}/rsshub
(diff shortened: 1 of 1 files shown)


TonyRL:
```diff
jobs:
  release:
+   if: github.repository_owner == 'DIYgod'
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v2
```

AgFlore:
I would rather recommend check whether ${{ secrets.DOCKER_USERNAME }} and ${{ secrets.DOCKER_PASSWORD }} is specified. 

There're users who want to build their own docker images for deployments, and we should not bring unnecessary extra trouble to them.

liyishuai:
> I would rather recommend check whether ${{ secrets.DOCKER_USERNAME }} and ${{ secrets.DOCKER_PASSWORD }} is specified.
> 
> There're users who want to build their own docker images for deployments, and we should not bring unnecessary extra trouble to them.

Good idea! Could you instruct me how to write the conditions? The one I wrote didn't work (liyishuai@c2c718ca6c6f58691bcd275b05591c67e6a14df5):
> The workflow is not valid. .github/workflows/docker-release.yml (Line: 17, Col: 9): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.DOCKER_USERNAME && secrets.DOCKER_PASSWORD

AgFlore:
> > The workflow is not valid. .github/workflows/docker-release.yml (Line: 17, Col: 9): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.DOCKER_USERNAME && secrets.DOCKER_PASSWORD

Ah yes Github Actions does not natively support doing so. One needs to copy the secret to env first, like https://github.com/actions/runner/issues/520#issuecomment-907770967

liyishuai:
Success: https://github.com/liyishuai/RSSHub/actions/runs/1792454129

TonyRL:
Checking the existence of `secrets.DOCKER_USERNAME` seems already enough. For security reasons, I against the idea of putting password/token to envs.

.github/workflows/docker-release.yml
```diff
@@ -12,8 +12,20 @@ on:
...
 jobs:
+  check-env:
+    runs-on: ubuntu-latest
+    outputs:
+      check-docker: ${{ steps.check-docker.outputs.defined }}
+    steps:
+      - id: check-docker
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+        if: "${{ env.DOCKER_USERNAME != '' }}"
```
TonyRL, line 23:
```suggestion
        if: ${{ env.DOCKER_USERNAME != '' }}
```
I think quotes can be skipped

.github/workflows/docker-release.yml
```diff
@@ -12,8 +12,20 @@ on:
...
+        run: echo "::set-output name=defined::true"
   release:
     runs-on: ubuntu-latest
+    needs: [check-env]
```
TonyRL, line 27:
```suggestion
    needs: check-env
```
No need for array.

TonyRL:
Reverted.

`secrets.DOCKER_USERNAME` is set and still skipping build.

See https://github.com/DIYgod/RSSHub/actions/runs/1794535484

.github/workflows/docker-release.yml
```diff
@@ -12,8 +12,20 @@ on:
...
+    steps:
+      - id: check-docker
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+        if: ${{ env.DOCKER_USERNAME != '' }}
+        run: echo "::set-output name=defined::true"
   release:
     runs-on: ubuntu-latest
+    needs: check-env
+    if: needs.check-env.outputs.defined == 'true'
```
AgFlore, line 28: if: needs.check-env.outputs.check-docker == 'true'

AgFlore: should work like https://github.com/AgFlore/RSSHub/runs/5066207981 @TonyRL @liyishuai

PR #9025, opened 2022-02-08
Fix acwifi.net support

tylinux:

```routes
/acwifi/
```

- [ ] New Route
- [ ] Documentation
  - [ ] CN
  - [ ] EN
- [x] 全文获取 fulltext
  - [x] Use Cache
- [ ] 日期和时间 date and time
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [x] `Puppeteer`

### DIFF
--- a/lib/routes/acwifi/index.js
+++ b/lib/routes/acwifi/index.js
@@ -1,14 +1,20 @@
-const got = require('@/utils/got');
 const cheerio = require('cheerio');
 
 module.exports = async (ctx) => {
-    const baseUrl = 'https://www.acwifi.net/';
-    const response = await got({
-        method: 'get',
-        url: baseUrl,
-    });
-    const $ = cheerio.load(response.data);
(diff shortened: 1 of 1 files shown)


TonyRL:
Did you try `https://www.acwifi.net/feed` before switching to puppeteer approach?

tylinux:
> Did you try `https://www.acwifi.net/feed` before switching to puppeteer approach?

Yes, I can comfirm it's worked.

TonyRL:
> > Did you try `https://www.acwifi.net/feed` before switching to puppeteer approach?
> 
> Yes, I can comfirm it's worked.

If `https://www.acwifi.net/feed` works then, you should use that instead of puppeteer approach

tylinux:
acwifi.net use baidu CDN to protect his resources, so this fix is unstable, close it.

PR #9213, opened 2022-02-27
feat(route)(telegram/channel): major rewrite

Rongronggg9:

https://github.com/DIYgod/RSSHub/issues/4777#issuecomment-628828048

```routes
/telegram/channel/jia0_sub
/telegram/channel/capoo_sub
/telegram/channel/NewlearnerChannel
/telegram/channel/NewlearnerChannel/Pixel
/telegram/channel/NewlearnerChannel/searchQuery=Pixel
/telegram/channel/awesomeDIYgod
/telegram/channel/awesomeDIYgod/%23DIYgod的豆瓣动态
/telegram/channel/awesomeDIYgod/searchQuery=%23DIYgod的豆瓣动态
/telegram/channel/PubTChannel
/telegram/channel/PubTChannel/showLinkPreview=1&showViaBot=1&showReplyTo=1&showFwdFrom=1&showFwdFromAuthor=1&showInlineButtons=1&showMediaTagInTitle=1&showMediaTagAsEmoji=1&includeFwd=1&includeReply=1&includeServiceMsg=1&includeUnsupportedMsg=1
/telegram/channel/PubTChannel/showLinkPreview=0&showViaBot=1&showReplyTo=1&showFwdFrom=1&showFwdFromAuthor=1&showInlineButtons=1&showMediaTagInTitle=1&showMediaTagAsEmoji=0&includeFwd=1&includeReply=1&includeServiceMsg=1&includeUnsupportedMsg=1
/telegram/channel/PubTChannel/showLinkPreview=1&showViaBot=1&showReplyTo=1&showFwdFrom=1&showFwdFromAuthor=0&showInlineButtons=0&showMediaTagInTitle=1&showMediaTagAsEmoji=0&includeFwd=1&includeReply=1&includeServiceMsg=0&includeUnsupportedMsg=0
/telegram/channel/PubTChannel/showLinkPreview=0&showViaBot=0&showReplyTo=0&showFwdFrom=0&showFwdFromAuthor=0&showInlineButtons=0&showMediaTagInTitle=0&showMediaTagAsEmoji=1&includeFwd=0&includeReply=0&includeServiceMsg=0&includeUnsupportedMsg=0
```

- [ ] 新的路由 New Route
- [x] 文档说明 Documentation
  - [x] 中文文档 CN
  - [x] 英文文档 EN
- [x] 全文获取 fulltext
  - [x] 使用缓存 Use Cache
  - [x] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
```
feat(route)(telegram/channel): major rewrite

feed title changed to English
route path changed
exclude 'no posts found' messages (always)

custom route params:
exclude forwarded/reply/service messages (selectable)
exclude unsupported messages (selectable, default)
remove link preview (selectable)
remove metadata (via bot / forwarded from / reply to) (selectable)
add forwarded message author (selectable, default)
add inline buttons (selectable)
remove media tags in title (selectable)
non-emoji media tags (selectable)
search query (selectable, retain backward compatibility)

message types newly supported / enhanced:
reply (previous not in media tags, lacks of via bot)
forwarded (previously not in media tags, lacks of via bot)
service (previous not in media tags, img not extracted)
via bot (previous unrecognized, treated as normal messages)
video (previous use another emoji as media tag)
GIF (previous unrecognized, treated as video)
poll (previous only poll title)
voice (previous use another emoji as media tag, no title/content)
music (previous unrecognized, treated as document)
document (previous no content, broken if multiple exist)
location (previous only media tag)
contact (previous totally unsupported)
sticker (previous only media tag)
animated sticker (previous totally unsupported)
unsupported (previous totally unsupported)

misc:
fix emoji in posts: weird format (Italic and Bold)
fix title: lacks of spaces
fix title: use pubDate even if media tags exist
fix broken documents message if multiple documents exist
fix mistaken exception description
fix media sequence: video always at last, https://github.com/DIYgod/RSSHub/issues/4777
fix: autofilled title (pubDate) not in UTC
use cache
minor refactor

Signed-off-by: Rongrong <15956627+Rongronggg9@users.noreply.github.com>
```

### DIFF
--- a/lib/v2/telegram/channel.js
+++ b/lib/v2/telegram/channel.js
@@ -1,23 +1,104 @@
+const config = require('@/config').value;
 const got = require('@/utils/got');
 const cheerio = require('cheerio');
 const { parseDate } = require('@/utils/parse-date');
 const { art } = require('@/utils/render');
 const path = require('path');
+const querystring = require('querystring');
+const { fallback, queryToBoolean } = require('@/utils/readable-social');
+
+/* message types */
+const REPLY = 'REPLY';
+const FORWARDED = 'FORWARDED';
+const SERVICE = 'SERVICE';
+const VIA_BOT = 'VIA_BOT';
+const VIDEO = 'VIDEO';
+const GIF = 'GIF';
+const PHOTO = 'PHOTO';
+const POLL = 'POLL';
+const VOICE = 'VOICE';
+const MUSIC = 'MUSIC';
+const DOCUMENT = 'DOCUMENT';
+const LOCATION = 'LOCATION';
+const CONTACT = 'CONTACT';
+const STICKER = 'STICKER';
+const ANIMATED_STICKER = 'ANIMATED_STICKER';
+// const VIDEO_STICKER = 'VIDEO_STICKER';  // not supported yet by t.me
+const UNSUPPORTED = 'UNSUPPORTED';
+
+/* message media tag dict */
+const mediaTagDict = {
+    REPLY: ['[Reply]', '↩️'],
+    FORWARDED: ['[Forwarded]', '🔁'],
+    SERVICE: ['[Service]', '🔧'],
+    VIA_BOT: ['[Via bot]', '🤖'],
+    VIDEO: ['[Video]', '🎬'],
+    GIF: ['[GIF]', '[GIF]'],
+    PHOTO: ['[Photo]', '🖼'],
+    POLL: ['[Poll]', '📊'],
+    VOICE: ['[Voice]', '🎙'],
+    MUSIC: ['[Music]', '🎵'],
+    DOCUMENT: ['[Document]', '📄'],
+    LOCATION: ['[Location]', '📍'],
+    CONTACT: ['[Contact]', '📱'],
+    STICKER: ['[Sticker]', '[Sticker]'],
+    ANIMATED_STICKER: ['[Animated Sticker]', '[Animated Sticker]'],
+    // VIDEO_STICKER: ['[Video Sticker]', '[Video Sticker]'],  // not supported yet by t.me
+    UNSUPPORTED: ['[Unsupported]', '🚫'],
+};
 
 module.exports = async (ctx) => {
     const username = ctx.params.username;
-    const searchQuery = ctx.params.searchQuery;
-    const resourceUrl = searchQuery ? `https://t.me/s/${username}?q=${encodeURIComponent(searchQuery)}` : `https://t.me/s/${username}`;
-
-    const { data } = await got.get(resourceUrl);
-    const $ = cheerio.load(data);
-    const list = $('.tgme_widget_message_wrap');
-
-    if (list.length === 0 && $('.tgme_channel_history').length === 0) {
-        throw `Unable to fetch message feed from this channel. Please check this URL to see if you can view the message preview: https://t.me/s/${username}`;
+    let routeParams = ctx.params.routeParams;
+    let showLinkPreview = true;
+    let showViaBot = true;
+    let showReplyTo = true;
+    let showFwdFrom = true;
+    let showFwdFromAuthor = true;
+    let showInlineButtons = false;
+    let showMediaTagInTitle = true;
+    let showMediaTagAsEmoji = true;
+    let includeFwd = true;
+    let includeReply = true;
+    let includeServiceMsg = true;
+    let includeUnsupportedMsg = false;
+    let searchQuery = routeParams; // for backward compatibility
+    if (routeParams && routeParams.search(/(^|&)(show(LinkPreview|ViaBot|ReplyTo|FwdFrom(Author)?|InlineButtons|MediaTag(InTitle|AsEmoji))|include(Fwd|Reply|(Service|Unsupported)Msg)|searchQuery)=/) !== -1) {
+        routeParams = querystring.parse(ctx.params.routeParams);
+        showLinkPreview = !!fallback(undefined, queryToBoolean(routeParams.showLinkPreview), showLinkPreview);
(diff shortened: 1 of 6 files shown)


TonyRL:
Please also take a look at deepscan issues

Rongronggg9:
> Please also take a look at deepscan issues

This should be a false positive:

```
Condition 'type !== FORWARDED' is always true.
```

```js
                    let mediaTag = '';
                    if (showMediaTagInTitle) {
                        msgTypes.forEach((type) => {
                            if (type !== REPLY || type !== FORWARDED || type !== VIA_BOT || (type === REPLY && showReplyTo) || (type === FORWARDED && showFwdFrom) || (type === VIA_BOT && showViaBot)) {
                                mediaTag += showMediaTagAsEmoji ? mediaTagDict[type][1] : mediaTagDict[type][0];
```

lib/v2/telegram/channel.js
```diff
@@ -3,29 +3,111 @@ const cheerio = require('cheerio');
...
 
-    const { data } = await got.get(resourceUrl);
+    const data = await ctx.cache.tryGet(
+        resourceUrl,
+        async () => {
+            const response = await got.get(resourceUrl);
+            return response.data;
+        },
+        300
+    ); // cache it for 5 minutes to avoid reload when routeParams changes
```
Rongronggg9, line 91: Can I safely use `config.cache.routeExpire` here? Since those who have their own instance may set a route expire time shorter than 5min.

TonyRL: If so, then those who want more frequent update on the content listing cannot benefit from it.
On the other hand, some instances(e.g., demo instance) may have a longer `routeExpire` value to prevent being throttled may also get into trouble.

Rongronggg9: > If so, then those who want more frequent update on the content listing cannot benefit from it.

I don't quite understand. It is `routeExpire` instead of `contentExpire`. If `routeParams` is unchanged, this cache never comes in handy because of the route cache hit. Only if `routeParams` changed, this cache can come in handy. 
Even if I don't apply a cache here, the problem is that, is there anyone who needs to benefit from changing `routeParams` to bypass the route cache?

> On the other hand, some instances(e.g., demo instance) may have a longer `routeExpire` value to prevent being throttled may also get into trouble.

Could you give an example of how this might happen? If `routeExpire` is more than 5min, then only cache it for 5min, would it be better?
My main concern is that if a lot of users subscribed to the same channel but applies different `routeParams`, it can cause a lot unnecessary duplicated web requests if cache not applied.

TonyRL: > It is `routeExpire` instead of `contentExpire`. If `routeParams` is unchanged, this cache never comes in handy because of the route cache hit.

If one set `config.cache.routeExpire` to 0 (completely disable route cache), with this settings, new feed items will delayed by 5 mins which does not seem alot. But I guess they set to 0 for instant notification and they have their reasons. Adding a fixed limit to it is not a good thing.

> Could you give an example of how this might happen?

<https://rsshub.app> has a `routeExpire` of 120mins.

> My main concern is that if a lot of users subscribed to the same channel but applied different routeParams, it can cause a lot of unnecessary duplicated web requests if cache is not applied.

I see your pov. `ctx.cache.tryGet()` already provide a cache time of 5 mins for key `resourceUrl` if no cache time provided and changing `routeParams` will not invalid caching according to your code.

> those who have their own instance may set a route expire time shorter than 5min.

If they set it shorter than 5 mins or even 0, they know the risks(generate lots of unnecessary traffics and getting ip banned) and benefits(quick update) of it.

Rongronggg9: > If one set `config.cache.routeExpire` to 0 (completely disable route cache), with this settings, new feed items will delayed by 5 mins which does not seem alot. But I guess they set to 0 for instant notification and they have their reasons. Adding a fixed limit to it is not a good thing.

Yeah, so I've asked if using `routeExpire` (instead of `contentExpire` by default, or fixed 5min) is a good idea.

> `ctx.cache.tryGet()` already provide a cache time of 5 mins for key `resourceUrl` if no cache time provided

5 mins? Doesn't it default to be `contentExpire`? The reason why I insist that here should not use default is that `contentExpire` is usually much too long. Either no caching is used, or use `routeExpire` as cache time. Using `contentExpire` as cache time is unbearable.

```js
tryGet: async (key, getValueFunc, maxAge = config.cache.contentExpire) => {
```

<hr>

Now I know what the problem really is. If I use `ctx.cache.tryGet()`, RSSHub will enforce refreshing the expiration time on each read. And this refresh will make provided `maxAge` useless — RSSHub always reset the expiration time to `config.cache.routeExpire`!
There is a workaround to make content cache behave like route cache — use `ctx.cache.get({refresh: false})` and `ctx.cache.set()` instead. What's your opinion about that?

I also wonder why `maxAge` is ineffective when refreshing, and why enforcing refresh on `ctx.cache.tryGet()`. If refreshing is selectable, those routes with tons of formatting options (e.g. Weibo, Twitter) can cache the initial entry too, just like route caching, making anti-rate-limit and web traffic saving more effective.

TonyRL: > There is a workaround to make content cache behave like route cache — `use ctx.cache.get({refresh: false})` and `ctx.cache.set()` instead. What's your opinion about that?

Sure, you can use it to cache the source page(with `refresh: false` on both set/get in this case).

> why enforcing refresh on `ctx.cache.tryGet()`.

Related https://github.com/DIYgod/RSSHub/issues/1182

notofoe:
Nice work, thank you!

PR #9264, opened 2022-03-07
ci: ignore route test on forks

liyishuai:
```routes
noroute
```
## 说明 / Note
Automated pull requests to forks like liyishuai/RSSHub#1007 are closed because `routes` was unspecified. No need to check this on forks.

### DIFF
--- a/.github/workflows/pr-deploy-route-test.yml
+++ b/.github/workflows/pr-deploy-route-test.yml
@@ -9,7 +9,7 @@ jobs:
   testRoute:
     name: Vercel Preview
     runs-on: ubuntu-latest
-    if: github.repository_owner == 'DIYgod'
+    if: github.actor != github.repository_owner
     steps:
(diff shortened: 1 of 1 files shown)


SukkaW (changes requested):
What happened if we decide to change our repo to a specific org like `RSSHub`? Depending on repo owner is a bit footgun. A better approach should be based on environment variable.

liyishuai:
Yes that makes sense. What non-private variable has this repo specified?

Rongronggg9:
what about `if: github.actor != github.repository_owner`?

liyishuai:
Nice catch. Thanks!

TonyRL:
FYI, you can disable unwanted workflow(s) in the tabs

liyishuai:
For future reference: The button is hidden under "..." menu. GitHub seems to have a different mindset from its users.
image

PR #9286, opened 2022-03-09
fix(core): non-collapsed feed fields content

Rongronggg9:

#9259

```routes
NOROUTE
```

FYI, the route in the screenshots below is `/amazon/ku/this`.

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
Unintended linebreaks, white spaces in some fields will be trimmed ([ref](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space)) by most RSS readers while rendering a post, but can break some RSS readers:
```xml
            <author><![CDATA[
      冯骥才       
       (作者) 
        ]]></author>
```

Thus, trimming them (just like `white-space: normal;`, but the spaces at the start/end are trimmed additionally) does not harm but can avoid many problems:
```xml
            <author><![CDATA[冯骥才 (作者)]]></author>
```

**What makes things worse is that if the whitespaces and linebreaks in the entry title uncollapsed, RSSHub will count those whitespaces and linebreaks during trimming the post title, making the post title even shorter than intended.**


lib/middleware/template.js
```diff
@@ -40,10 +40,22 @@ module.exports = async (ctx, next) => {
...
+                ctx.state.data.title = ctx.state.data.title.replace(/\s+/g, ' ').trim();
+            }
+
+            if (ctx.state.data.subtitle) {
+                ctx.state.data.subtitle = ctx.state.data.subtitle.replace(/\s+/g, ' ').trim();
+            }
+
+            if (ctx.state.data.author) {
+                ctx.state.data.author = ctx.state.data.author.replace(/\s+/g, ' ').trim();
+            }
```
SukkaW, line 53:
```suggestion
            const rWhiteSpace = /\s+/;
            const rAllWhiteSpace = /\s+/g;
            if (ctx.state.data.title && rWhiteSpace.test(ctx.state.data.title)) {
                ctx.state.data.title = ctx.state.data.title.replace(rAllWhiteSpace, ' ').trim();
            }
            if (ctx.state.data.subtitle && rWhiteSpace.test(ctx.state.data.subtitle)) {
                ctx.state.data.subtitle = ctx.state.data.subtitle.replace(rAllWhiteSpace, ' ').trim();
            }
            if (ctx.state.data.author && rWhiteSpace.test(ctx.state.data.author)) {
                ctx.state.data.author = ctx.state.data.author.replace(rAllWhiteSpace, ' ').trim();
            }
```

lib/middleware/template.js
```diff
@@ -40,10 +40,22 @@ module.exports = async (ctx, next) => {
...
+
             ctx.state.data.item &&
                 ctx.state.data.item.forEach((item) => {
                     if (item.title) {
```
SukkaW, line 57:
```suggestion
                    if (item.title && rWhiteSpace.test(item.title)) {
```

lib/middleware/template.js
```diff
@@ -40,10 +40,22 @@ module.exports = async (ctx, next) => {
...
-                        item.title = item.title.trim();
+                        item.title = item.title.replace(/\s+/g, ' ').trim();
```
SukkaW, line 58:
```suggestion
                        item.title = item.title.replace(rAllWhiteSpace, ' ').trim();
```

lib/middleware/template.js
```diff
@@ -54,6 +66,10 @@ module.exports = async (ctx, next) => {
                         }
                     }
 
+                    if (item.author) {
+                        item.author = item.author.replace(/\s+/g, ' ').trim();
+                    }
```
SukkaW, line 71:
```suggestion
                    if (item.author && rWhiteSpace.test(item.author)) {
                        item.author = item.author.replace(rAllWhiteSpace, ' ').trim();
                    }
```

SukkaW (changes requested):
This part of the middleware will be a hot path, so we'd better make sure it runs fast and not add too many overheads.

Besides my current suggestions, there are also more that can be done, e.g. by extracting a `collapseWhitespace` function, it enables v8 to optimize the call and further utilize the v8 inline cache.

Rongronggg9:
> extracting a `collapseWhitespace` function, it enables v8 to optimize the call and further utilize the v8 inline cache.

OK, I've done this. Please check if I made something wrong or not, since I am not so familiar with these speedup techniques.

lib/middleware/template.js
Rongronggg9: We should not do this, since a title trimming must happen even if there is no whitespace to be collapsed.

Rongronggg9:
> What makes things worse is that if the whitespaces and linebreaks in the entry title uncollapsed, RSSHub will count those whitespaces and linebreaks during trimming the post title, making the post title even shorter than intended.

Here is a severe example:
https://rsshub.app/newzmz
```
            <title><![CDATA[第四季合集                                
                                    了不起的麦瑟尔夫人 第四季
                                    The Marve...]]></title>
```
There are 32 whitespaces and 1 linebreak after `第四季合集`, 36 whitespaces before and 1 linebreak after `了不起的麦瑟尔夫人 第四季`, and 36 whitespaces before `The Marvelous Mrs. Maisel`, causing unintended title trimming severely.

SukkaW (approved):

PR #9395, opened 2022-03-26
refactor(core): unified memory cache

Rongronggg9:

Close #

```routes
NOROUTE
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
After bumping lru-cache to 7.x, we no longer need to maintain two LRU caches in the memory.

### DIFF
--- a/lib/middleware/cache/index.js
+++ b/lib/middleware/cache/index.js
@@ -27,10 +27,10 @@ if (config.cache.type === 'redis') {
     globalCache.set = cacheModule.set;
 } else if (config.cache.type === 'memory') {
     cacheModule = require('./memory');
-    const { pageCache } = cacheModule.clients;
+    const { memoryCache } = cacheModule.clients;
     globalCache.get = (key) => {
         if (key && cacheModule.status.available) {
-            return pageCache.get(key);
+            return memoryCache.get(key, { updateAgeOnGet: false });
         }
     };
     globalCache.set = (key, value, maxAge) => {
@@ -41,7 +41,7 @@ if (config.cache.type === 'redis') {
             value = JSON.stringify(value);
         }
         if (key) {
-            return pageCache.set(key, value, { ttl: maxAge * 1000 });
+            return memoryCache.set(key, value, { ttl: maxAge * 1000 });
         }
     };
 } else {
--- a/lib/middleware/cache/memory.js
+++ b/lib/middleware/cache/memory.js
@@ -3,15 +3,9 @@ const config = require('@/config').value;
 
 const status = { available: false };
 
-const pageCache = new Lru({
+const memoryCache = new Lru({
     ttl: config.cache.routeExpire * 1000,
     max: config.memory.max,
-});
-
-const routeCache = new Lru({
-    ttl: config.cache.routeExpire * 1000,
-    max: config.memory.max,
-    updateAgeOnGet: true,
 });
 
 status.available = true;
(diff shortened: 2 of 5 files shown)


Rongronggg9:
2ed9fd5c5ff235d37b6f0b31577573ae3b700add added a config `MEMORY_MAX`...
In fact, by just deleting the `max` parameter, the cache size can be set to infinite just like before.
However, if the cache size is limited, and the route cache and content cache share the same cache backend, the content cache may easily flush all the route cache away.

So @TonyRL, what is your opinion about that? I've come out four solutions:
1. Make the cache size unlimited again. (just like before)
2. Make the cache size unlimited again, meanwhile enable [`ttlAutopurge`](https://github.com/isaacs/node-lru-cache#options). (making memory cache behave nearly the same as Redis cache, but may significantly degrade performance)
3. Preserve `pageCache` used only for route cache, and making any route cache, no matter need or need not refreshing, go into `memoryCache`.
4. Increase the default value of `MEMORY_MAX` to a high enough number.

TonyRL:
Option 1:
`(node:77022) [LRU_CACHE_UNBOUNDED] UnboundedCacheWarning: TTL caching without ttlAutopurge, max, or maxSize can result in unbounded memory consumption.`
With `ENABLE_CLUSTER` enabled + `memory` caching option, RSSHub can be OOM-killed easily on low-cost hosting plaform.

Option 2:
No warnings but may significantly degrade performance as said in lru-cache docs.

Option 4:
Memory are pre-allocated in lru-cache v7 instead of growing in v6. I would prefer having a low value and setting it to a high value manually if necessary.

Option 3+5:
> Preserve pageCache used only for route cache, and making any route cache, no matter need or need not refreshing, go into memoryCache.

I think that's somehow what RSSHub is doing right now. One enforced cache for final RSS output, another optional cache for content or other things but doing it in two individual caches instead one.
Also, it has a cache keys of `koa-redis-cache:md5(ctx.request.path)` which I guess is quite less like to have a collision.

Rongronggg9:
> I think that's somehow what RSSHub is doing right now. 

No, setting content cache with `{ refresh: false }` makes it go into `pageCache`, otherwise `routeCache`. This behavior significantly brings two disadvantages:
1. Setting a lot of cache entries with `{ refresh: false }` may flush all the route cache away. Though currently no routes were doing this, there do be necessary to make those routes with tons of formatting settings cache their initial entry with `{ maxAge: config.cache.routeExpire, refresh: false }` to avoid being anti-crawled as already discussed in https://github.com/DIYgod/RSSHub/pull/9213#discussion_r816702782. Once this PR is merged, it will start doing this.
2. When getting or setting content cache, an identical value of `refresh` is strictly needed. This is a really weird behavior because setting a cache always "refreshes" something. No one will think of this unless they have already been told or carefully read the code, however, the documents never mention this point. The best example is that both you (@TonyRL) and I, never noticed that I had [forgotten to pass the `refresh` parameter to `cache.set` in the enhanced `cache.tryGet`](https://github.com/DIYgod/RSSHub/pull/9227/files#diff-a4a41994eac1c095beaf1d25fff1a6baad1b8ee32947122862ef5f892c5bb390R62). (I apologize for my negligence...)

So since other solutions are not acceptable, I will implement Solution 3.

TonyRL:
I would prefer [3292af2](https://github.com/DIYgod/RSSHub/commit/3292af2e4ca16a7b59a49548a02f048bacdcd8e8) to further shrink memory usage(2 * `MEMORY_MAX` vs 1)
I also didn't notice lru-cache can do `get(key, { updateAgeOnGet, allowStale } = {}) => value` in v7 until this PR comes in.

Rongronggg9:
> I would prefer 3292af2 to further shrink memory usage(2 * `MEMORY_MAX` vs 1)

I see. Though 2 * `MEMORY_MAX` is exactly current behavior, I will follow your advice and reset this PR to 3292af2. 

<details>
<summary>off-topic content</summary>

> I also didn't notice lru-cache can do get(key, { updateAgeOnGet, allowStale } = {}) => value in v7 until this PR comes in.

I've noticed it before the migration to v7. At that time I'd planned to create a migration PR and implement a unified memory cache, but you finished the migration soon. You are absolutely a diligent and nice maintainer.

</details>

PR #9466, opened 2022-04-04
docs: disable actions on forks

liyishuai:
Follow up #9264 

```routes
NOROUTE
```


TonyRL:
I don’t really see the necessary of this. 
When someone fork a repository with workflows on GitHub, workflows aren’t enabled by default.
Apart from that, there’s is a huge warning that tells people think twice before enabling workflows.

liyishuai:
I see. That makes sense.

TonyRL:
Most of the GitHub Apps("Apps" from https://github.com/apps) does not need GitHub Actions to work with.

PR #9487, opened 2022-04-06
feat(utils): add utils for WeChat MP

Rongronggg9:

Close #

```routes
/wechat/tgchannel/lifeweek
/wechat/feeds/Mzg2ODAyNTgyMQ==
/wechat/mp/homepage/MzA3MDM3NjE5NQ==/16
/wechat/mp/msgalbum/MzA3MDM3NjE5NQ==/1375870284640911361
/gzh360/category
/gzh360/gzh/北京青年报
/wechat/feeddd/6131e1441269c358aa0e2141
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [x] 全文获取 fulltext
  - [x] 使用缓存 Use Cache
  - [x] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
```
feat(*): add utils for WeChat MP

Motivation:
There are multiple routes that need to fetch articles from WeChat MP.
However, letting them fetch articles by themselves could potentially
lead to cache key collisions. Even if cache key collisions do not occur,
un-normalized URL could potentially lead to duplicated requests.
What's more, articles from WeChat MP have weird formats and need to be
fixed. Creating a universal function to do this work can create some
ease for new route contributors.

Note:
In order to make this PR atomic as much as possible, I did not touch
those broken routes. Once this PR is merged, I will try to fix them.

Signed-off-by: Rongrong <15956627+Rongronggg9@users.noreply.github.com>
```

### DIFF
--- a/lib/utils/wechat-mp.js
+++ b/lib/utils/wechat-mp.js
@@ -1,0 +1,164 @@
+/**
+ * Author: @Rongronggg9
+ *
+ * There are at least three folders which are relevant with WeChat MP (Official Account Platform / Media Platform):
+ * lib/route/tencent/wechat
+ * lib/v2/wechat
+ * lib/v2/gzh360
+ *
+ * If your new route is not in the above folders, please add it to the list.
+ *
+ * If your route need to fetch MP articles from mp.weixin.qq.com, you SHOULD use `finishArticleItem`.
+ * However, if your route need to determine some metadata by itself, you MAY use `fetchArticle`.
+ * If you find more metadata on the webpage, consider modifying `fetchArticle` to include them.
+ * NEVER fetch MP articles from mp.weixin.qq.com in your route in order to avoid cache key collision.
+ * NO NEED TO use cache if you are using `finishArticleItem` or `fetchArticle`, they will handle cache for you.
+ *
+ * If your route fetch MP articles from other websites, you SHOULD use `fixArticleContent` to fix the content format.
+ * If you find more fixes that should be applied, consider modifying `fixArticleContent` to include them.
+ *
+ * For more details of these functions, please refer to the jsDoc in the source code.
+ */
+
+const got = require('@/utils/got');
+const cheerio = require('cheerio');
+const { parseDate } = require('@/utils/parse-date');
+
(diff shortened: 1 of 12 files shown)


TonyRL (commented):
It would be better if the annotations of the new utility can follow jsdoc

lib/utils/wechat-mp.js
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Fetch article and its metadata from WeChat MP (mp.weixin.qq.com), then fill the `item` object with the result.
+ *
+ * If you use this function, no need to call `fetchArticle` or `fixArticleContent`
+ *
+ * A new route SHOULD use this function instead of manually calling the above functions
+ *
+ * An existing route adopting this function SHOULD either:
+ * - set `skipLink` to true (not recommended)
+ * - set `item.guid` to `item.link` BEFORE calling this function
```
TonyRL, line 134: Could you explain why it's necessary to set `item.guid` to `item.link`? Middleware will handle it if no guid is set and guid isn't used in this utility.

https://github.com/DIYgod/RSSHub/blob/e0b7ca676d8b5817b474861ad674205444f3a15e/lib/views/rss.art#L35
https://github.com/DIYgod/RSSHub/blob/e0b7ca676d8b5817b474861ad674205444f3a15e/lib/views/atom.art#L37

Rongronggg9: Only existing routes need to do that in order not to break existing subscriptions.

Different guid means different posts. RSS readers use guid to uniquely distinguish posts. A new guid will make the RSS reader consider the post as a never-seen post. Thus, it is necessary to make sure not to change the guid of the post of existing routes.

Rongronggg9: This utility will normalize the URL. If not manually setting guid to the original URL, the guid will be changed and make RSS readers consider an ever-seen post as never-seen.

TonyRL: Understood.

PR #9611, opened 2022-04-25
feat(route): add chinafactcheck

kdanfly:

Close #

```routes
/chinafactcheck
```

- [x] 新的路由 New Route
- [x] 文档说明 Documentation
  - [x] 中文文档 CN
  - [ ] 英文文档 EN
- [x] 全文获取 fulltext
  - [x] 使用缓存 Use Cache
  - [x] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note

有据|国际新闻事实核查


TonyRL:
FYI, WordPress-powered site has its RSS feed under `/feed` and REST API under `/wp-json` endpoint

kdanfly:
> FYI, WordPress-powered site has its RSS feed under `/feed` and REST API under `/wp-json` endpoint

Gotcha! My mistake, but the radar does not recognize at the same time. This site is missing the `link` type of `application/rss+xml` to show it's feed link.

kdanfly:
Please check `https://rsshub.app/chinafactcheck`

It's not working without `/` ending, and there is no cache this time. Maybe route under `v2` mast have `/` ending.

TonyRL:
>  and there is no cache this time

That's not true. Maybe I didn't explain it clearly last time:

First, rsshub.app consists of multiple instances and their code are passively synced with GitHub hourly. Meanwhile, docs.rsshub.app is in sync with GitHub utilising the `push` event and built within 5 mins right after `push` is triggered.

Second, demo instances use both Redis and Cloudflare for caching. You can confirm this by
```shell
curl -I https://rsshub.app/chinafactcheck
```

Third, on docs.rsshub.app, you can see a `status up` or a `status down` badge. This is done by checking the response code from the defined example using rsshub.app in real-time (Note that Cloudflare will cache the response from rsshub.app including 404s).

Fourth, there be a notification when a new route is pushed in [@awesomeRSSHub](https://t.me/awesomeRSSHub) Telegram Channel which people will visit the docs for new routes.

As a result, new routes(which should follow v2 standard) like `/chinafactcheck` do not work on demo instances right after it comes out. Adding a `/` will work because it's an unseen request for Cloudflare.

PR #9683, opened 2022-05-03
chore(*): migrate to GHA-based PR route test

Rongronggg9:

Close #

```routes
NOROUTE
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
To prove it works properly, check https://github.com/Rongronggg9/RSSHub/pull/160 and https://github.com/Rongronggg9/RSSHub/pull/159

In order to test with ease, you may create PR in my fork.

### DIFF
--- a/scripts/workflow/test-route/test.js
+++ b/scripts/workflow/test-route/test.js
@@ -1,6 +1,6 @@
 /* eslint-disable */
 
-module.exports = async ({ github, context, core }, baseUrl, routes, number) => {
+module.exports = async ({ github, context, core, got }, baseUrl, routes, number) => {
     if (routes[0] === 'NOROUTE') {
         return;
     }
@@ -15,27 +15,26 @@ module.exports = async ({ github, context, core }, baseUrl, routes, number) => {
     for (const lks of links) {
         core.info(`testing route:  ${lks}`);
         // Intended, one at a time
-        const res = await github.request(`GET ${lks}`).catch((err) => {
+        const res = await got(lks).catch((err) => {
             com += `
-
 <details>
-    <summary><a href="${lks}">${lks}</a>  - **Failed**</summary>
+    <summary><a href="${lks}">${lks}</a> - <b>Failed</b></summary>
 
 \`\`\`
-    ${err}
+${err}
 \`\`\`
 </details>
 
 `;
         });
-        if (res && res.data) {
-            const { data } = res;
+        if (res && res.body) {
+            const { body } = res;
             com += `
 <details>
-    <summary><a href="${lks}">${lks}</a>  - Success</summary>
+    <summary><a href="${lks}">${lks}</a> - Success</summary>
 
-\`\`\`
-    ${data.split('\n').slice(0, 30).join('\n')}
+\`\`\`rss
+${body.replace(/\s+(\n|$)/g, '\n')}
 \`\`\`
 </details>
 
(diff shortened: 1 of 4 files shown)


scripts/workflow/test-route/test.js
```diff
@@ -15,27 +15,26 @@ module.exports = async ({ github, context, core }, baseUrl, routes, number) => {
...
-            const { data } = res;
+        if (res && res.body) {
+            const { body } = res;
             com += `
 <details>
-    <summary><a href="${lks}">${lks}</a>  - Success</summary>
+    <summary><a href="${lks}">${lks}</a> - Success</summary>
 
-\`\`\`
-    ${data.split('\n').slice(0, 30).join('\n')}
```
TonyRL, line 38: Just a reminder
> PR body/Issue comments are still stored in MySQL as a mediumblob with a maximum value length of 262,144. This equals a limit of 65,536 4-byte unicode characters.

https://github.community/t/148867/2

Rongronggg9: This is indeed a problem, but there's just not a better solution to store the result. Or alternatively, would you be willing to create an orphan branch to temporarily store them (meanwhile clean outdated ones regularly)? There are a lot of workflow actions that can achieve the goal.

TonyRL: > a better solution to store the result

What about using workflow artifacts? Just like 
https://github.com/DIYgod/RSSHub/blob/225d922d9cc9b52035509aace0f12ca3767f74df/.github/workflows/docker-test.yml#L65-L70

Rongronggg9: I don't think it is an auditor-friendly solution. All files are always packed into a single uncompressed zip file (artifact). To fetch a single file, one needs to download the whole artifact and unpack it. If an auditor needed to do so, why not just pull the PR and test locally? 🤔 ~Is there an API to directly get a single file in it?~ For now, it seems impossible to even download a single file, not to mention preview it online. Ref: https://github.community/t/serving-html-artifacts/159407/5, https://github.com/actions/upload-artifact/issues/14

The reason why I suggest creating a new orphan branch to store them is that only in this solution, the file can be viewed online, and meanwhile, no external dependencies are introduced (otherwise we need some online services like Pastebin).

But I consider it is not so urgent to fix this since more-then-2-byte Unicode characters are rare. Mostly we can assume that the average character length is less than 2 bytes. Thus the length limit is close to up to 13.1k characters.

Another solution is to simply automatically split a comment into multiple ones.

TonyRL: Then I think we can leave it as it is unless there's pr that has tons of lengthy articles trying to break this limit.

Rongronggg9: Fine. Let's wait for good news from https://github.com/github/roadmap/issues/470 to see if it helps. If so, I will try to implement a better solution.

TonyRL: > HttpError: Validation Failed: {"resource":"IssueComment","code":"unprocessable","field":"data","message":"Body is too long (maximum is 65536 characters)"}

https://github.com/DIYgod/RSSHub/actions/runs/2275206140

PR #9796, opened 2022-05-21
fix(core/parameter): absurd wrong path cache time

Rongronggg9:

Close #

```routes
NOROUTE
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
```js
    if (!ctx.state.data && !ctx._matchedRoute) {
        // Given that the official demo has a cache TTL of 2h, a "wrong path" page will be cached by CloudFlare for
        // 200h (8.33d).
        // What makes it worse is that the documentation contains status badges to detect the availability of routes,
        // but the documentation is updated more timely than the official demo, so the every example path of every
        // new route will probably have a "wrong path" page cached for at least 200h soon after accepted. That is to
        // say, the example paths of a new route will probably be unavailable on the public demo in the first 200h
        // after accepted.
        // As a conclusion, the next 3 lines has been commented out. (exactly the same behavior as any internal error)
        // ctx.set({
        //     'Cache-Control': `public, max-age=${config.cache.routeExpire * 100}`,
        // });
        throw Error('wrong path');
    }
```

### DIFF
--- a/lib/middleware/parameter.js
+++ b/lib/middleware/parameter.js
@@ -10,9 +10,17 @@ module.exports = async (ctx, next) => {
     await next();
 
     if (!ctx.state.data && !ctx._matchedRoute) {
-        ctx.set({
-            'Cache-Control': `public, max-age=${config.cache.routeExpire * 100}`,
-        });
+        // Given that the official demo has a cache TTL of 2h, a "wrong path" page will be cached by CloudFlare for
+        // 200h (8.33d).
+        // What makes it worse is that the documentation contains status badges to detect the availability of routes,
+        // but the documentation is updated more timely than the official demo, so the every example path of every
+        // new route will probably have a "wrong path" page cached for at least 200h soon after accepted. That is to
+        // say, the example paths of a new route will probably be unavailable on the public demo in the first 200h
+        // after accepted.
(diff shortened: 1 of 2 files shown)


TonyRL:
I think it would be better to reduce the multiplier, let's say, 3x, instead of completely removed it.

Rongronggg9:
Yeah, I did have thought the same at the beginning but after a deep dive, I doubted that the purpose of multiplying the cache time is fairly unclear. I mean, I didn't see any fundamental differences between a "wrong path" error and other internal errors, from routes or core, in terms of server loads or attack surface. Self-consistent is important. If the former statement is true, no reason to "privilege" "wrong path" errors.

TonyRL:
So the limit was introduced back in the days when there was a limited budget for servers and there were plenty of requests for deleted routes. Both of them are no longer the case for today.

Rongronggg9:
Thanks for your answer. That answered my question in mind.

PR #9970, opened 2022-06-17
feat(fix): use `list = await Promise.all(...)` instead of `await Promise.all(...)`; 

Derekmini:

Close #

```routes
/elsevier/signal-processing/192
/elsevier/signal-processing
/heu/gx/list/xw/yw
/heu/job/list/tzgg
/heu/uae/list/3751
/heu/yjsy/list/2981
/scitation/aapt/ajp
/scitation/aapt/ajp/COMPUTATIONAL+PHYSICS
/springer/journal/10450
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [x] 全文获取 fulltext
  - [x] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

### DIFF
--- a/lib/v2/elsevier/issue.js
+++ b/lib/v2/elsevier/issue.js
@@ -17,7 +17,7 @@ module.exports = async (ctx) => {
     });
     const $ = cheerio.load(response.data);
     const jrnlName = $('.anchor.js-title-link').text();
-    const list = $('.js-article')
+    let list = $('.js-article')
         .map((_, item) => {
             const title = $(item).find('.js-article-title').text();
             const authors = $(item).find('.js-article__item__authors').text();
@@ -37,7 +37,7 @@ module.exports = async (ctx) => {
         art(path.join(__dirname, 'templates/description.art'), {
             item,
         });
-    await Promise.all(
+    list = await Promise.all(
         list.map((item) =>
             ctx.cache.tryGet(item.link, async () => {
                 const response2 = await got(`${host}/science/article/pii/${item.id}`, {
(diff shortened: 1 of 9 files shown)


lib/v2/elsevier/issue.js
```diff
@@ -17,7 +17,7 @@ module.exports = async (ctx) => {
     });
     const $ = cheerio.load(response.data);
     const jrnlName = $('.anchor.js-title-link').text();
-    const list = $('.js-article')
+    let list = $('.js-article')
```
SukkaW, line 20:
```suggestion
    const items = $('.js-article')
```

lib/v2/elsevier/issue.js
```diff
@@ -37,7 +37,7 @@ module.exports = async (ctx) => {
         art(path.join(__dirname, 'templates/description.art'), {
             item,
         });
-    await Promise.all(
+    list = await Promise.all(
```
SukkaW, line 40:
```suggestion
    const list = await Promise.all(
```

SukkaW (changes requested):

Derekmini:
@SukkaW 

Do you advise me to use "const" instead of "let" in the script? 

Since the script is generated by my own RSSHub template, it may be used for routes that use or don't use the "await Promise.all" function. As a result, I use "let" in the route that uses "await Promise.all" and "const" in the route that doesn't.

In order to maintain consistency, I want to make those contractions uniform. Therefore, may I discard your changes?

PR #10714, opened 2022-09-06
feat(core): support proxy config for pupperteer

DIYgod:

Close #

```routes
NOROUTE
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note

This is a core featrue

### DIFF
--- a/lib/utils/puppeteer.js
+++ b/lib/utils/puppeteer.js
@@ -14,11 +14,6 @@ if (config.proxyUri && typeof config.proxyUri === 'string') {
 } else if (config.proxy && config.proxy.protocol && config.proxy.host && config.proxy.port) {
     proxyUri = `${config.proxy.protocol}://${config.proxy.host}:${config.proxy.port}`;
 }
-if (proxyUri) {
-    proxyChain.anonymizeProxy(proxyUri).then((anonymizedProxyUri) => {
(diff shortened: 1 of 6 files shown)


lib/utils/puppeteer.js
```diff
@@ -54,6 +49,9 @@ module.exports = async (extraOptions = {}) => {
         puppeteer.use(require('puppeteer-extra-plugin-stealth')());
     }
     let browser;
+    if (proxyUri) {
+        options.args.push(`--proxy-server=${await proxyChain.anonymizeProxy(proxyUri)}`);
+    }
```
DIYgod, line 54: Why has this code been moved here, I think it would be more efficient to put it up there and only execute it once 👀

TonyRL: It can't pass this simple test https://github.com/DIYgod/RSSHub/blob/9329a44c80a37b9aec18a233c9e3bdb3f1c7621d/test/utils/puppeteer.js#L83-L90
The launch options observed from puppeteer process are still the `options` defined above.

PR #10744, opened 2022-09-08
feat(core): 增加 `sorted` 参数来控制结果排序

### LINKED ISSUE #10743: 增加部分 Feed 不按 pubDate 排序的「通用参数」：sorted
### 这是一个什么样的功能？

在[通用参数](https://github.com/DIYgod/RSSHub/blob/master/docs/parameter.md)中添加新参数 `sorted`，默认为 `true`。

当 `sorted=false`，[则不对返回的内容按发布时间进行排序](https://github.com/DIYgod/RSSHub/blob/master/lib/middleware/parameter.js#L57)。

### 这个功能可以解决什么问题？

部分网站，如[南京大学本科生院教学信息网](https://jw.nju.edu.cn/ggtz/list1.htm)，会将某些条目（通常为长期有效或较为重要的通知）置顶。通过 `sorted` 选项，用户可以保留这些置顶的行为。

### 额外描述

晚些时候将会打开相应的 Pull Request 提供试验性的实现。

### 这不是重复的功能请求和 RSS 提案

- [X] 我已经搜索了 [现有 issue](https://github.com/DIYgod/RSSHub/issues)，以确保这项功能尚未被请求及不是 [新的 RSS 提案](https://github.com/DIYgod/RSSHub/issues/new?assignees=&labels=RSS+proposal&template=rss_request_zh.yml)。

cqjjjzr:

Close #10743

```routes
NOROUTE
```

## 说明 / Note

参照 Issue #10743 。

### DIFF
--- a/docs/en/parameter.md
+++ b/docs/en/parameter.md
@@ -80,6 +80,12 @@ Set `limit` to limit the number of articles in the feed.
 
 E.g. Dribbble Popular Top 10 [https://rsshub.app/dribbble/popular?limit=10](https://rsshub.app/dribbble/popular?limit=10)
 
+## Sorted
+
+Set `sorted` to control whether to sort the output by the publish date (`pubDate`). This is useful for some feeds that pin some entries at the top. Default to `true` i.e. the output is sorted.
+
+E.g. NJU Undergraduate Bulletin Board <https://rsshub.app/nju/jw/ggtz?sorted=false>
+
 ## Fulltext
 
 Enable fulltext via `mode` parameter.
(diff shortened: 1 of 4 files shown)


TonyRL:
FYI, most RSS readers will sort feed items according to their pubDate by default. Apart from that, once a feed item has been indexed, RSS readers will not notify users about this item again, so the order does not really matter. (I guess users won't read feed output from RSS directly.)

cqjjjzr:
At least for Thunderbird which I use, you can optionally select to sort based on the order the entries are received. And the order does matter for certain feeds like the one I mentioned in #10743 .

Apart from that, IMHO giving users an option to have the feeds represented in the order from the upstream is reasonable: the different order would be a little confusing.

PR #10920, opened 2022-09-27
feat(route): add Yoasobi Official support

Kiotlin:

Close #

```routes
/yoasobi-music/info/news
/yoasobi-music/info/biography
/yoasobi-music/live
/yoasobi-music/media
```

- [x] 新的路由 New Route
- [x] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [x] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
Yoasobi, a Japanese music [superduo](https://en.wikipedia.org/wiki/Superduo) formed by Sony Music. Add new routes for its [official website](https://www.yoasobi-music.jp/).


lib/v2/yoasobi-music/info.js
```diff
@@ -0,0 +1,64 @@
...
+            // the article content
+            description: art(path.join(__dirname, 'templates/info.art'), {
+                image: i.image,
+                category: i.category,
+                description: i.description.replace(/\n/g, '<br>'),
+            }),
+            // the article publish time
+            pubDate: parseDate(i.date),
+            // the article link
+            link: i.id ? `${officialUrl}/${i.id}` : officialUrl,
```
TonyRL, line 61: link is also used as `guid` if not specified, which should be unique

lib/v2/yoasobi-music/templates/live.art
```diff
@@ -0,0 +1,12 @@
+{{ if image }}
+    <img src={{image}} />
+{{ /if }}
+{{@ description }}
+{{ each sessions se i}}
```
TonyRL, line 5:
```suggestion
{{ each sessions se }}
```
You can omit `i` if it isn't used

lib/v2/yoasobi-music/media.js
```diff
@@ -0,0 +1,56 @@
...
+            // the article title
+            title: i.title,
+            // the article content
+            description: art(path.join(__dirname, 'templates/media.art'), {
+                date: i.date,
+                weekDay: i.weekDay,
+                postFix: i.startTime && i.endTime && i.tvStation ? `${i.startTime} ~ ${i.endTime} ${i.tvStation}` : null,
+                description: i.description,
+            }),
+            // no pubDate
```
TonyRL, line 51: `i.date` can be used as pubDate

lib/v2/yoasobi-music/media.js
```diff
@@ -0,0 +1,56 @@
...
+            // the article link
+            link: officialUrl,
```
TonyRL, line 53: link is also used as `guid` if not specified, which should be unique

lib/v2/yoasobi-music/media.js
Kiotlin: Actually the `i.date` refers to the air time of a tv/radio/... show, which has a different meaning with the date a post being published. Is there any side effect if I left `pubDate` param empty?

lib/v2/yoasobi-music/info.js
Kiotlin: Same with [media.js](https://github.com/DIYgod/RSSHub/pull/10920#discussion_r981154009). Every news have their own unique page url `${officialUrl}/${i.id}`, but when comes to `biography`, they become pure `<div>` embeded in the page. [news](https://www.yoasobi-music.jp/news) [biography](https://www.yoasobi-music.jp/biography)

lib/v2/yoasobi-music/media.js
TonyRL: explicitly set `guid` to something like `guid: i.title + i.date`

lib/v2/yoasobi-music/media.js
TonyRL: You can leave `pubDate` empty and the feed will still work. But the user side, users cannot sort their feeds by date if it isn't provided

Kiotlin: OK, I see. I will use `i.date` as pubDate

PR #11038, opened 2022-10-09
feat: Add(route) 北深

whuzjw:

Close #

```routes
/pkuszcs/xwzx/0
/pkuszcs/xwzx/1
/pkuszcs/xwzx/2
```

- [x] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [x] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
北深cs的rss, 文档暂时未写

### DIFF
--- a/lib/v2/pkuszcs/router.js
+++ b/lib/v2/pkuszcs/router.js
@@ -1,0 +1,3 @@
+module.exports = function (router) {
+    router.get('/xwzx/:type', require('./xwzx'));
+};
--- a/lib/v2/pkuszcs/xwzx.js
+++ b/lib/v2/pkuszcs/xwzx.js
@@ -9,7 +9,7 @@ const map = new Map([
 ]);
 
 module.exports = async (ctx) => {
-	let type = Number.parseInt(ctx.params.type);
+	const type = Number.parseInt(ctx.params.type);
 
 	if (type !== 0) {
 		const id = map.get(type).link;
@@ -41,7 +41,7 @@ module.exports = async (ctx) => {
     	};
 	}
 	else {
-		let all = new Array();
+		const all = new Array();
 		/* eslint-disable no-await-in-loop */
 		for (let i = 1; i < 3; i++) {
 			const id = map.get(i).link;
@@ -54,7 +54,7 @@ module.exports = async (ctx) => {
 
 			const $ = cheerio.load(data);
 			const list = $('.list_box_sousuo');
-    		let items = list &&
+    		const items = list &&
         	list
             	.map((index, item) => {
                 	item = $(item);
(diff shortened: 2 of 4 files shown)


lib/v2/pkuszcs/router.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function (router) {
+    router.get('/xwzx/:type', require('./xwzx'));
```
TonyRL, line 2: Please use the namespace `pku` instead of `pkuszcs` since that's the right domain name

lib/v2/pkuszcs/xwzx.js
```diff
@@ -0,0 +1,84 @@
...
+module.exports = async (ctx) => {
+	const type = Number.parseInt(ctx.params.type);
+
+	if (type !== 0) {
+		const id = map.get(type).link;
+		// 发起 HTTP GET 请求
+		const response = await got({
+	    	method: 'get',
+	    	url: String(id),
+		});
```
TonyRL, line 20: got request can be simplified
```suggestion
		const response = await got(String(id));
```

lib/v2/pkuszcs/xwzx.js
```diff
@@ -0,0 +1,84 @@
...
+		const $ = cheerio.load(data);
+		const list = $('.list_box_sousuo');
+    	const items = list &&
+        	list
+            	.map((index, item) => {
+                	item = $(item);
+                	return {
+                    	title: item.find('a').attr('title'),
+                    	link: item.find('a').attr('href'),
+		    	pubDate: date(item.find('span').text()),
```
TonyRL, line 32: Please use `parseDate()`
docs: https://docs.rsshub.app/joinus/pub-date.html

lib/v2/pkuszcs/xwzx.js
```diff
@@ -0,0 +1,84 @@
...
+    	ctx.state.data = {
+        	title: map.get(type).title + ' - pkuszcs',
+        	link: map.get(type).link,
+        	item: items,
+    	};
+	}
+	else {
+		const all = new Array();
+		/* eslint-disable no-await-in-loop */
+		for (let i = 1; i < 3; i++) {
```
TonyRL, line 46: Is it really necessary to run in sequential order?
I guess you can use [`Promise.all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) instead

TonyRL (changes requested):
Please double-check your code. It should not return only 1 item.

PR #10189, opened 2022-07-10
fix(route): add Open Graph image to GitHub trending

jameschensmith:
## 该 PR 相关 Issue / Involved issue

Not applicable.

## 完整路由地址 / Example for the proposed route(s)

```routes
/github/trending/daily/any/en
```

## 新 RSS 检查列表 / New RSS Script Checklist
  
- [ ] 新的路由 New Route
  - [x] 跟随 [v2 路由规范](https://docs.rsshub.app/joinus/script-standard.html) Follows [v2 Script Standard](https://docs.rsshub.app/en/joinus/script-standard.html)
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
- [ ] 反爬/频率限制 anti-bot or rate limit?
  - [ ] 如果有, 是否有对应的措施? If yes, do your code reflect this sign?
- [ ] [日期和时间](https://docs.rsshub.app/joinus/pub-date.html) [date and time](https://docs.rsshub.app/en/joinus/pub-date.html)
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] 添加了新的包 New package added
- [ ] `Puppeteer`

## 说明 / Note

Adds Open Graph image for each repository to the GitHub trending feed. Also, corrects unknown languages. I'd love to use GraphQL in place of got / cheerio, but I [ran into an issue where `openGraphImageUrl` is not returning what we want here](https://support.github.com/ticket/personal/0/1700375). Hopefully that gets fixed, and we can improve this in a future pull request 🙂 

This is my first pull request here. I couldn't understand most of the checklist options, but if I'm missing anything please let me know and I'll try to get it corrected 😊👍

PR #11188, opened 2022-10-31
fix(route): use GraphQL query in GitHub trending

jameschensmith:

Not applicable.

```routes
/github/trending/daily/any/en
/github/trending/weekly/javascript
/github/trending/monthly/unknown
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note

Follow-up from #10189. GitHub now returns `openGraphImageUrl` correctly. Uses GraphQL to get desired repository data instead of attempting to query it from the repository page.

### DIFF
--- a/docs/en/programming.md
+++ b/docs/en/programming.md
@@ -137,7 +137,7 @@ GitHub provides some official RSS feeds:
 
 ### Trending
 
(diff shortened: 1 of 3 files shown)


TonyRL (commented):
Please add `selfhost="1"` attribute to the docs since it requires users to setup github token after this change

TonyRL:
> does `selfhost="1"` mean that it can't be called from the main instance?

Yes. However, some of the `Self-host only` routes are working on demo instance because they have their tokens configured.

j1g5awi:
Hey, guys? what's problem of you? 
Why make it can not work without github_token? GitHub Trending RSS just works well even without github_token.

PR #11241, opened 2022-11-08
feat(route): Add ymgal.games RSS

SunBK201:

Close #

```routes
/ymgal/article/
/ymgal/article/news
/ymgal/article/column
```

- [x] 新的路由 New Route
- [x] 文档说明 Documentation
  - [x] 中文文档 CN
  - [ ] 英文文档 EN
- [x] 全文获取 fulltext
  - [x] 使用缓存 Use Cache
  - [x] 可以解析 Parsed
  - [x] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

### DIFF
--- a/lib/v2/ymgal/article.js
+++ b/lib/v2/ymgal/article.js
@@ -1,0 +1,47 @@
+const got = require('@/utils/got');
+const cheerio = require('cheerio');
+
+const host = 'https://www.ymgal.games';
+
+const types = {
+    news: '?type=NEWS&page=1',
+    column: '?type=COLUMN&page=1',
+};
+
+module.exports = async (ctx) => {
+    const type = ctx.params.type || 'news';
+    const link = `${host}/co/topic/list` + types[type];
+
+    const response = await got(link);
+    const data = response.data.data;
+
+    const items = await Promise.all(
+        data.map(async (item) => {
+            const itemUrl = host + '/co/article/' + item.topicId;
+            const description = await ctx.cache.tryGet(item.topicId, async () => {
+                const result = await got.get(itemUrl);
+                const $ = cheerio.load(result.data);
+                return $('article').html().trim();
+            });
+            const single = {
+                title: item.title,
+                link: itemUrl,
+                pubDate: new Date(item.publishTime).toUTCString(),
+                description,
+            };
+            return Promise.resolve(single);
+        })
+    );
+
+    let info = '资讯';
+    if (type === 'column') {
+        info = '专栏';
+    }
+
+    ctx.state.data = {
+        title: `月幕 Galgame - ${info}`,
+        link: 'https://www.ymgal.games/co/article',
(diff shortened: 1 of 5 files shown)


lib/v2/ymgal/article.js
```diff
@@ -0,0 +1,47 @@
...
+                const $ = cheerio.load(result.data);
+                return $('article').html().trim();
+            });
+            const single = {
+                title: item.title,
+                link: itemUrl,
+                pubDate: new Date(item.publishTime).toUTCString(),
+                description,
+            };
+            return Promise.resolve(single);
```
TonyRL, line 32:
```suggestion
        data.map((item) => {
            const itemUrl = host + '/co/article/' + item.topicId;
            return ctx.cache.tryGet(itemUrl, async () => {
                const result = await got(itemUrl);
                const $ = cheerio.load(result.data);
                const description = $('article').html().trim();
                return {
                    title: item.title,
                    link: itemUrl,
                    pubDate: timezone(parseDate(item.publishTime), 8),
                    description,
                };
            });
```
- You can use `itemUrl` as the cache key instead of `topicId` since it has a lower chance of collision and provides more details for those managing Redis.
- You can cache the title, link and pubDate as an object as well.
- The `publishTime` provided from the API may need an offset after comparing with the article

lib/v2/ymgal/maintainer.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    '/article': ['SunBK201'],
```
TonyRL, line 2:
```suggestion
    '/article/:type?': ['SunBK201'],
```
The key here is the same as the path in `router.js`

lib/v2/ymgal/radar.js
```diff
@@ -0,0 +1,19 @@
...
+                title: '资讯',
+                docs: 'https://docs.rsshub.app/anime.html#yue-mu-galgame',
+                source: ['/'],
+                target: '/ymgal/article/news',
+            },
+            {
+                title: '专栏',
+                docs: 'https://docs.rsshub.app/anime.html#yue-mu-galgame',
+                source: ['/'],
+                target: '/ymgal/article/co',
```
TonyRL, line 15: I think this should be column according to the docs
```suggestion
                target: '/ymgal/article/column',
```

lib/v2/ymgal/article.js
```diff
@@ -0,0 +1,61 @@
...
+                const response = await got(`${host}/co/topic/list${type}`);
+                data.push.apply(data, response.data.data);
+            })
+        );
+    } else {
+        const response = await got(link);
+        data = response.data.data;
+    }
+
+    data = data.slice(0, 10);
```
TonyRL, line 30:
```suggestion
        );
        data = data.sort((a, b) => b.publishTime - a.publishTime).slice(0, 10);
    } else {
        const response = await got(link);
        data = response.data.data;
    }

```

PR #11242, opened 2022-11-08
fix(route): nautilus

emdoe:

Close #

```routes
/nautil/topic/arts
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
thanks a lot for the previous feedbacks.


docs/en/new-media.md
```diff
@@ -514,9 +514,9 @@ Edition
 
 ### Topics
 
-<RouteEn author="emdoe" example="/nautilus/topic/arts" path="/nautilus/topic/:tid" :paramsDesc="['topic']">
+<RouteEn author="emdoe" example="/nautil/topic/arts" path="/nautil/topic/:tid" :paramsDesc="['topic']">
```
emdoe, line 517: I would suggest continuing to use `nautilus` rather than `nautil`

TonyRL: I know this is somehow a breaking change to users.
However, this helps reducing some unnecessary namespace collisions using purely the domain name instead of alias/site name, imagine someone wants an RSS feed for https://nautilus.org

emdoe: to me it's perfectly fine; however if users want feeds for _all_ three of these websites, the issue would persist. but thanks for the help!

TonyRL (commented):
Turn out this is a WordPress-based site. Using its REST API saves a lot of requests (2 vs 21 default) and efforts.

1. obtain the topic id list from the API and save it to cache
```js
    const categoryIdMap = await ctx.cache.tryGet('nautil:categories', async () => {
        const { data } = await got(`${baseUrl}/wp-json/wp/v2/categories`, {
            searchParams: {
                per_page: 100,
            },
        });
        return data.map((item) => ({
            id: item.id,
            name: item.name,
            slug: item.slug,
        }));
    });
```

2. obtain the articles
```js
    const { data: list } = await got(`${baseUrl}/wp-json/wp/v2/posts`, {
        searchParams: {
            categories: categoryIdMap.find((item) => item.slug === ctx.params.tid.toLowerCase()).id,
            per_page: ctx.query.limit ? parseInt(ctx.query.limit) : 20,
        },
    });
```

3. prepare the output
```js
    const out = list.map((item) => {
        const head = item.yoast_head_json;
        const $ = cheerio.load(item.content.rendered, null, false);
        // lazyload images
        $('img').each((_, e) => {
            e = $(e);
            e.attr('src', e.attr('data-src') ?? e.attr('srcset'));
            e.attr('src', e.attr('src').split('?')[0]);
            e.removeAttr('data-src');
            e.removeAttr('srcset');
        });
        return {
            title: item.title.rendered,
            author: item.yoast_head_json.author,
            description: art(path.join(__dirname, 'templates/description.art'), {
                head,
                rendered: $.html(),
            }),
            link: item.link,
            pubDate: parseDate(item.date_gmt),
        };
    });
```
```html
{{ if head.og_image }}
{{ each head.og_image img }}
<img src="{{ img.url.split('?')[0] }}">
{{ /each }}
{{ /if }}
{{@ rendered }}
```

4. some extra libs and slight tweaks
```js
const { parseDate } = require('@/utils/parse-date');
const { art } = require('@/utils/render');
const path = require('path');
const baseUrl = 'https://nautil.us';
```
```js
    ctx.state.data = {
        title: 'Nautilus | ' + categoryIdMap.find((item) => item.slug === ctx.params.tid.toLowerCase()).name,
        link: `${baseUrl}/topics/${ctx.params.tid}/`,
        item: out,
    };
```

PR #11619, opened 2023-01-13
fix(route): bloomberg author

josh:

Follow up #11615

```routes
/bloomberg/authors/ARbTQlRLRjE/matthew-s-levine
/bloomberg/authors/AS7Hj1mBMGM/mark-gurman
```

- [ ] 新的路由 New Route
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note

I've notice the official feeds are often out of date or missing items that are present on the source HTML page. This switches over to extract the article links directly from the markup. The downside being this requests being more prone to hitting a captcha. Are there any captcha workarounds RSSHub uses that might be useful here? Thanks for reviewing all these PRs so fast! Hopefully others find these routes useful.


lib/v2/bloomberg/authors.js
```diff
@@ -1,14 +1,37 @@
...
+            })
+            .get();
+    });
 
 module.exports = async (ctx) => {
     const { id, slug } = ctx.params;
-    const feed = await parser.parseURL(`https://www.bloomberg.com/authors/${id}/${slug}.rss`);
-    const item = await asyncPoolAll(1, feed.items, (item) => parseArticle(item, ctx));
+    const link = `https://www.bloomberg.com/authors/${id}/${slug}`;
+    const list = await parseArticleList(link, ctx);
```
TonyRL, line 29: Caching this list is not recommended as new items may take way longer to appear

TonyRL (commented):
I think they used IP-based throttling. I tried switching UAs and cookies but doesn't work well.

josh:
Pushed commits for your suggestions.

> I think they used IP-based throttling. I tried switching UAs and cookies but doesn't work well.

What do you think about first trying to fetch from the html page, else falling back to the current feed approach if it's blocked? Not sure if any other routes follow this pattern. But it could help reliability.

bigfei:
Bloomberg uses [PX3](https://www.perimeterx.com) to protect their sites, it is not easy to bypass the PX3 captcha. I suggest giving the user the right to choose the data from either RSS reed or HTML page by a param like 'authors/ARbTQlRLRjE/matthew-s-levine?from=rss/site'.

lib/v2/bloomberg/authors.js
josh: So I did remove the cache, but now I'm thinking it still might be a good idea to avoid hitting the website too often to trigger the captcha. How long is the cache TTL? These specific pages don't update to often, so even once 24h would be okay.

josh:
@bigfei that's a good idea too. I saw you implemented most of the initial bloomberg scraper. Nice work! Especially the article extraction stuff. I see alot of it uses a secret api. Did you remember if there was anything to get articles by author via that approach?

lib/v2/bloomberg/authors.js
TonyRL: You may try this instead.
```js
const config = require('@/config').value;

const res = await ctx.cache.tryGet(key, returnValueFn(), config.cache.routeExpire, false);
```
`routeExpire` is set to 2 hrs on demo instance.
By setting the 4th parameter to `false`, the cache won't renew its expiry when `key` is hit.

josh:
Pushed up some more changes based on feedback. Thanks for the suggestions!

Added back some caching to the HTML fetch approach. Restored old rss feed method as an option. It can be explicitly requested with `?source=rss` or used when the html fetch fails.

bigfei:
@josh For authors, there seems to have an API: `/lineup/api/lazy_load_author_stories?slug=AS7Hj1mBMGM/mark-gurman&authorType=default&page=1` can be used as the source. Then extract the article link by XPath `a.story-list-story__image` to the response of that API. There is no need to get from RSS to HTML anymore if you can call that real-time API. 

> @bigfei that's a good idea too. I saw you implemented most of the initial bloomberg scraper. Nice work! Especially the article extraction stuff. I see alot of it uses a secret api. Did you remember if there was anything to get articles by author via that approach?

josh:
@bigfei great suggestion! I pushed some changes to pull from that api endpoint.

josh:
> http://localhost:1200/bloomberg/authors/AS7Hj1mBMGM/mark-gurman - Failed

Ugh, it appears that API route still might get blocked sometimes.

bigfei:
@TonyRL plz help to review this pull request.

PR #11740, opened 2023-01-31
feat(route): add support for XAUFE jiaowu route

shaokeyibb:

Close #

```routes
/xaufe/jiaowu/tzgg
```

- [x] 新的路由 New Route
- [x] 文档说明 Documentation
  - [x] 中文文档 CN
  - [ ] 英文文档 EN
- [x] 全文获取 fulltext
  - [x] 使用缓存 Use Cache
  - [x] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

### DIFF
--- a/lib/v2/xaufe/jiaowu.js
+++ b/lib/v2/xaufe/jiaowu.js
@@ -28,10 +28,10 @@ module.exports = async (ctx) => {
 
     const $ = cheerio.load(response);
 
-    const data = $('ul', '.main_conRCb')
-        .children('li')
+    const data = $('.main_conRCb ul li')
         .slice(0, 20)
-        .map((_, item) => {
+        .toArray()
+        .map((item) => {
             item = $(item);
             const pubDate = item.children('span').text();
             const title = item.find('a em').text();
@@ -44,8 +44,7 @@ module.exports = async (ctx) => {
                 title,
                 link,
             };
-        })
-        .get();
+        });
 
     ctx.state.data = {
         title: `${category.title}-${rootMeta.title}`,
@@ -63,13 +62,9 @@ module.exports = async (ctx) => {
                             })
                         ).body
                     );
-                    return {
-                        title: item.title,
-                        author: /作者：(\S*)\s{4}/g.exec($('p', '.main_contit').text())[1],
-                        description: $('#vsb_content').html(),
-                        pubDate: item.pubDate,
-                        link: item.link,
-                    };
+                    item.author = /作者：(\S*)\s{4}/g.exec($('p', '.main_contit').text())[1];
+                    item.description = $('#vsb_content').html();
+                    return item;
                 })
             )
         ),
--- a/lib/v2/xaufe/jiaowu.js
+++ b/lib/v2/xaufe/jiaowu.js
@@ -35,7 +35,10 @@ module.exports = async (ctx) => {
             item = $(item);
             const pubDate = item.children('span').text();
             const title = item.find('a em').text();
-            const link = item.children('a').attr('href').replace('../', rootMeta.url);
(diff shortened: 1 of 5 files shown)


lib/v2/xaufe/jiaowu.js
```diff
@@ -0,0 +1,77 @@
...
+        await got({
+            method: 'get',
+            url: rootMeta.url + category.url,
+        })
+    ).body;
+
+    const $ = cheerio.load(response);
+
+    const data = $('ul', '.main_conRCb')
+        .children('li')
```
TonyRL, line 32:
```suggestion
    const data = $('.main_conRCb ul li')
```

lib/v2/xaufe/jiaowu.js
```diff
@@ -0,0 +1,77 @@
...
+        .slice(0, 20)
```
TonyRL, line 33: The list has only 16 items

lib/v2/xaufe/jiaowu.js
```diff
@@ -0,0 +1,77 @@
...
+                .children('a')
+                .attr('href')
+                .replace(/\.\.\//g, rootMeta.url);
+            return {
+                pubDate: parseDate(pubDate),
+                title,
+                link,
+            };
+        })
+        .get();
```
TonyRL, line 48:
```suggestion
        .toArray()
        .map((item) => {
            item = $(item);
            const pubDate = item.children('span').text();
            const title = item.find('a em').text();
            const link = item
                .children('a')
                .attr('href')
                .replace(/\.\.\//g, rootMeta.url);
            return {
                pubDate: parseDate(pubDate),
                title,
                link,
            };
        });
```

lib/v2/xaufe/jiaowu.js
```diff
@@ -0,0 +1,77 @@
...
+
+    ctx.state.data = {
+        title: `${category.title}-${rootMeta.title}`,
```
TonyRL, line 51: I believe you can obtain the same value using $('head title').text()

lib/v2/xaufe/jiaowu.js
```diff
@@ -0,0 +1,77 @@
...
+                            })
+                        ).body
+                    );
+                    return {
+                        title: item.title,
+                        author: /作者：(\S*)\s{4}/g.exec($('p', '.main_contit').text())[1],
+                        description: $('#vsb_content').html(),
+                        pubDate: item.pubDate,
+                        link: item.link,
+                    };
```
TonyRL, line 72: You can reuse `item` in `map` instead of creating a new array
```suggestion
                    item.author = /作者：(\S*)\s{4}/g.exec($('p', '.main_contit').text())[1];
                    item.description = $('#vsb_content').html();
                    return item;
```

lib/v2/xaufe/jiaowu.js
shaokeyibb: Yep, but there's some other website may apply here in the future, they may not equal to the HTML title xD

lib/v2/xaufe/jiaowu.js
TonyRL: This helps save some memories since `Array.prototype.map()` will [create a new array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)

shaokeyibb: > This helps save some memories since `Array.prototype.map()` will [create a new array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)

Got it.

PR #11953, opened 2023-02-25
fix: Improve route "codeforces" time expression readability

Bubbu0129:

```routes
/codeforces/contests
```

## 说明 / Note
The time was previously presented bare and coarse. This PR makes it more human-readable.
Plus, `relativeTime` was removed, as it only depends on when the endpoint is requested.

### DIFF
--- a/lib/v2/codeforces/contests.js
+++ b/lib/v2/codeforces/contests.js
@@ -2,15 +2,14 @@ const got = require('@/utils/got');
 const path = require('path');
 const { art } = require('@/utils/render');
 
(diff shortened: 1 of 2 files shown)


TonyRL (commented):
Please use `dayjs` instead
docs: https://day.js.org/en/

Bubbu0129:
Where? On startTime or durationTime?

TonyRL:
Both, including `sec2str`

Bubbu0129:
Legit?

PR #12151, opened 2023-03-21
docs: use WHATWG URL

x2cf:

```routes
NOROUTE
```

## 说明 / Note

`url.resolve` is deprecated.


TonyRL:
Quick start has been completely rewritten.

PR #12152, opened 2023-03-21
fix(route): fix the error of route hk01/channel when channel does not have publishName

### LINKED ISSUE #12141: 希望可以添加香港01即时路由
### 这是一个什么样的功能？

image

### 这个功能可以解决什么问题？

此路由可以包含网站一切即时发布的新闻和文章

### 额外描述

_No response_

### 这不是重复的功能请求和 RSS 提案

- [X] 我已经搜索了 [现有 issue](https://github.com/DIYgod/RSSHub/issues)，以确保这项功能尚未被请求及不是 [新的 RSS 提案](https://github.com/DIYgod/RSSHub/issues/new?assignees=&labels=RSS+proposal&template=rss_request_zh.yml)。

5upernova-heng:

Close #12141 

```routes
/hk01/channel/1
/hk01/latest
```

- [x] 新的路由 New Route
- [x] 文档说明 Documentation
  - [x] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
In #12141 "即时" can be fetched by api `https://web-data.api.hk01.com/v2/feed/category/:id`, which is already implemented by route `/hk01/channel/:id`.

However, categories(or channels) like "即时" does not have `category` object in its json data. This will cause an error when using it's property to be the title of the rss feed. For example `/hk01/channel/0`. This commit is to fix this problem.

What's more, the only way I've found to get rss subscription like "即时", is to look up what the api the page called, which might be hard for those users who are not familiar with web to find the `id` of the page they want. It will be better if there's a proper solution to this problem.

### DIFF
--- a/docs/new-media.md
+++ b/docs/new-media.md
@@ -4181,6 +4181,32 @@ wechat-feeds 来源[已停止更新](https://github.com/hellodword/wechat-feeds/
 
 </Route>
 
+## 香港 01
+
+### 热门
+
+<Route author="hoilc Fatpandac nczitzk" example="/hk01/hot" path="/hk01/hot" radar="1" rssbud="1"/>
+
+### 栏目
+
+<Route author="hoilc Fatpandac nczitzk" example="/hk01/zone/11" path="/hk01/zone/:id" :paramsDesc="['栏目 id, 可在 URL 中找到']" radar="1" rssbud="1"/>
+
+### 子栏目
+
+<Route author="hoilc Fatpandac nczitzk" example="/hk01/channel/391" path="/hk01/channel/:id" :paramsDesc="['子栏目 id, 可在 URL 中找到']" radar="1" rssbud="1"/>
+
+### 专题
+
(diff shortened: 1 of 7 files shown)


TonyRL (changes requested):
This does not close #12141 at all. #12141 uses a different API than `lib/v2/hk01/channel.js`

5upernova-heng:
> This does not close #12141 at all. #12141 uses a different API than `lib/v2/hk01/channel.js`

You are right. #12141 uses a different API. Should I open other PR to do this, or just in this one?

TonyRL:
Simply push everything to this PR is enough.

5upernova-heng:
It seems that `whatsmyua.info` website is down right now(used in Jest test). How can I rerun the test later?

PR #12205, opened 2023-03-30
fix(route): fix gov/sichuan/deyang/govpulicinfo

zytomorrow:

Close #

```routes
/gov/sichuan/deyang/govpublicinfo/绵竹市
```

- [ ] 新的路由 New Route
- [x] 文档说明 Documentation
  - [x] 中文文档 CN
  - [ ] 英文文档 EN
- [x] 全文获取 fulltext
  - [x] 使用缓存 Use Cache
  - [x] 可以解析 Parsed
  - [x] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
因网址调整，修复信息获取
image
image


lib/v2/gov/sichuan/deyang/govpublicinfo.js
```diff
@@ -0,0 +1,115 @@
...
+            },
+        },
+    },
+};
+
+const getInfoUrlList = async (rootUrl, infoBasicUrl) => {
+    const infoUrlList = [];
+    const expectToday = new Date();
+    let notTodayCount = 0;
+    for (let pageNum = 1; ; pageNum++) {
```
TonyRL, line 29: Use the `while` loop if you don't want the middle expression of `for` loop

lib/v2/gov/sichuan/deyang/govpublicinfo.js
```diff
@@ -0,0 +1,115 @@
...
+                source: $('#symbol > div:nth-child(2) > div:nth-child(2)').text().split('：')[1].trim(),
+                content: $('#container > div.ewb-white > div.ewb-article-detail').html(),
+                file: fileList,
+                link: item.url,
+                _isCompleteInfo: true,
+            };
+        } catch (err) {
+            return {
+                title: item.title,
+                pubDate: Date(),
```
TonyRL, line 90: Don't add a date if the site does not provide one.

lib/v2/gov/sichuan/deyang/govpublicinfo.js
```diff
@@ -0,0 +1,116 @@
...
+        if (infoList.length) {
+            for (const infoItem of infoList) {
+                const date = parseDate(timezone($('span', infoItem).html()), +8);
+                // 判断当前信息日期是否是今日
+                if (date.getUTCFullYear() === expectToday.getUTCFullYear() && date.getUTCMonth() === expectToday.getUTCMonth() && date.getUTCDate() === expectToday.getUTCDate()) {
+                    infoUrlList.push({
+                        title: $('a', infoItem).attr('title'),
+                        url: `${rootUrl}${$('a', infoItem).attr('href')}`,
+                    });
+                } else {
```
TonyRL, line 46: But why push the entry only if it's today?
Users won't get any items on 2/4 if all items are from 1/4.

zytomorrow: I know this problem.Always, today's news are needed.
If has no limit,I don't hava a better way to handle so many news when first request this router with no limit params,and in this case, caches will become so big.
can you have some  advise, i can change this.
thx.

TonyRL: I think scraping only the first page is fine IMO. It will only send 16 requests on the first RSS generation and should occupy ~4MB of cache.

zytomorrow: I have modified it as you said.
Thx.

lib/v2/gov/sichuan/deyang/govpublicinfo.js
```diff
@@ -0,0 +1,97 @@
...
+    // 非当前日期文章计数，部分旧文章可能会置顶，目前为发现置顶数超过10
+    const infoList = $('body > div.container > div.ewb-white > div.ewb-job > ul > li');
+    // 判断当前页是否有数据
+    for (const infoItem of infoList) {
+        infoItemList.push({
+            title: $('a', infoItem).attr('title'),
+            url: `${rootUrl}${$('a', infoItem).attr('href')}`,
+        });
+    }
+    return infoItemList;
```
TonyRL, line 38:
```suggestion
    const response = await got(infoBasicUrl);
    const $ = cheerio.load(response.data);
    // 非当前日期文章计数，部分旧文章可能会置顶，目前为发现置顶数超过10
    const infoList = $('body > div.container > div.ewb-white > div.ewb-job > ul > li').toArray().map((item) => ({
        title: $('a', item).attr('title'),
        url: `${rootUrl}${$('a', item).attr('href')}`,
    }));
    return infoList;
```

PR #12422, opened 2023-05-02
fix(route): iwara subscription get token

FeCCC:

Close #

```routes
/iwara/subscriptions
```

- [ ] 新的路由 New Route
- [x] 文档说明 Documentation
  - [x] 中文文档 CN
  - [x] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note
The new version of the iwara website uses token, so that when there is no token, the subscribed content cannot be obtained, and what is obtained is the latest content on the website. It is now modified to use the username and password to log in to obtain the token, and then use the api to obtain the content

### DIFF
--- a/lib/v2/iwara/subscriptions.js
+++ b/lib/v2/iwara/subscriptions.js
@@ -8,30 +8,31 @@ const md = require('markdown-it')({
 });
 
 module.exports = async (ctx) => {
+    if (!config.iwara || !config.iwara.username || !config.iwara.password) {
+        throw Error('Iwara subscription RSS is disabled due to the lack of <a href="https://docs.rsshub.app/install/#pei-zhi-bu-fen-rss-mo-kuai-pei-zhi">relevant config</a>');
+    }
+
     const rootUrl = `https://www.iwara.tv`;
     const username = config.iwara.username;
     const password = config.iwara.password;
 
-    if (username === undefined || password === undefined) {
-        throw Error('Iwara subscription RSS is disabled due to the lack of <a href="https://docs.rsshub.app/install/#pei-zhi-bu-fen-rss-mo-kuai-pei-zhi">relevant config</a>');
-    }
-
     // get token
-    const loginResponse = await got({
-        method: 'post',
-        url: 'https://api.iwara.tv/user/login',
-        headers: {
-            'content-type': 'application/json',
-        },
-        data: JSON.stringify({
-            email: username,
-            password,
-        }),
-    });
-
-    const authHeaders = {
-        authorization: 'Bearer ' + loginResponse.data.token,
-    };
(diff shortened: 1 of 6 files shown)


lib/v2/iwara/subscriptions.js
```diff
@@ -9,39 +9,63 @@ const md = require('markdown-it')({
...
-    const videoSubUrl = 'https://api.iwara.tv/videos?page=0&limit=30&subscribed=true';
-    const imageSubUrl = 'https://api.iwara.tv/images?page=0&limit=30&subscribed=true';
-    const cookie = config.iwara.cookie;
+    const username = config.iwara.username;
+    const password = config.iwara.password;
 
-    if (cookie === undefined) {
+    if (username === undefined || password === undefined) {
         throw Error('Iwara subscription RSS is disabled due to the lack of <a href="https://docs.rsshub.app/install/#pei-zhi-bu-fen-rss-mo-kuai-pei-zhi">relevant config</a>');
     }
```
TonyRL, line 17: https://github.com/DIYgod/RSSHub/blob/6e26d8c8cda0d3d5cbaa97d8008602ecbd64aa89/lib/v2/discord/channel.js#L8-L10
Similarly
```js
if (!config.iwara || !username || !password) {
```

lib/v2/iwara/subscriptions.js
```diff
@@ -9,39 +9,63 @@ const md = require('markdown-it')({
...
+        },
+        data: JSON.stringify({
+            email: username,
+            password,
+        }),
+    });
+
+    const authHeaders = {
+        authorization: 'Bearer ' + loginResponse.data.token,
+    };
```
TonyRL, line 34: Please cache the authorization header using a cache key like `iwara:token` so we don't need to login again.

lib/v2/iwara/subscriptions.js
```diff
@@ -8,40 +8,70 @@ const md = require('markdown-it')({
...
+                },
+                data: JSON.stringify({
+                    email: username,
+                    password,
+                }),
+            });
+            return {
+                authorization: 'Bearer ' + loginResponse.data.token,
+            };
+        },
```
TonyRL, line 37: After some testing with the site, the token you returned here is `refresh_token` which will be expired after 1 month. However, this `refresh_token` is used to generate a short-lived `access_token` from https://api.iwara.tv/user/token.

Apart from that, requests for both `videoSubUrl` and `imageSubUrl` are using `access_token` instead of `refresh_token`.

PR #12672, opened 2023-06-16
fix(route): allow Mastodon without extra config

### LINKED ISSUE #11836: /mastodon/acct: Get the feed directly from the acct's instance, configuration-free
### What feature is it?

For Mastodon instances with deployment version 4.0.0 or above, calling the `search` API no longer requires a user access token by default, so when fetching a user's tweet, RSSHub may now first try to call the `search` API of the instance they are on, get the user ID, and then fetch posts directly from that instance, possibly eliminating the need to fetch on other instance given by RSSHub provider.

For example, to fetch posts from [@CatWhitney@mastodon.social](https://mastodon.social/@CatWhitney), if https://mastodon.social is deployed with version 4.0.0 or above, RSSHub can now fetch them directly from https://mastodon.social. Fetching posts from the instance given by RSSHub provider may be not necessary. 

### What problem does this feature solve?

If some users are on a newer version of Mastodon's instance, fetching posts from these users via the route `/mastodon/acct/:acct/statuses/:only_media?` no longer requires additional configuration of the RSSHub.

### Additional description

For backward compatibility, the original query mechanism must remain in place as a failback. The server where the queried user is on does not have to be Mastodon.

Links to check: 
 - <https://docs.joinmastodon.org/methods/search/>
 - <https://docs-develop.pleroma.social/backend/development/API/differences_in_mastoapi_responses/>

### This is not a duplicated feature request or new RSS proposal

- [X] I have searched [existing issues](https://github.com/DIYgod/RSSHub/issues) to ensure this feature has not already been requested and this is not a [new RSS proposal](https://github.com/DIYgod/RSSHub/issues/new?assignees=&labels=RSS+proposal&template=rss_request_en.yml).

myl7:

Close #11836

```routes
/mastodon/acct/CatWhitney@mastodon.social/statuses
```

- [ ] 新的路由 New Route
- [x] 文档说明 Documentation
  - [x] 中文文档 CN
  - [x] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] `Puppeteer`

## 说明 / Note

After Mastodon 4.0.0, `search` API does not always require an access token, at least for the use in RSSHub.
Also, the acct domain and API host can be inferred from the passed acct if the Mastodon server has the same `LOCAL_DOMAIN` and `WEB_DOMAIN`.
The config is still kept in case users need to override them.
More details are in the corresponding feature proposal issue.


TonyRL (commented):
Please update the docs as well.

myl7:
Corresponding docs have been added

PR #12932, opened 2023-08-04
docs: fix text for the language dropdown

ghost:

```routes
NOROUTE
```


TonyRL (commented):
They are inverted by design. Consider a person who doesn't know Chinese and visited https://docs.rsshub.app. How can one change the site language to English if the site shows `选择语言` instead of `Languages`?

PR #13144, opened 2023-08-28
fix: V2EX Route Exclude Duplicate Posts

0xleizhang:

Close #

```routes
/v2ex/tab/hot?limit=50&filter_title=remote
/v2ex/tab/jobs?limit=50&filter_title=remote
```
 

## Note / 说明

v2ex的帖子链接是变化的 例如：` /t/968950#reply4` 会自动添加上最新的回帖ID，这样会导致rss出现多个相同帖子的结果。
这个PR将#及之后的内容除去了，同时优化了limit由固定值10改成优先从参数取值。

The post link on v2ex is variable, for example: `/t/968950#reply4` will automatically add the latest reply ID, which causes multiple identical posts to appear in the RSS feed.
This PR removes everything after "#" in the link and optimizes the limit value by prioritizing it from the parameter instead of using a fixed value of 10.

image


TonyRL (commented):
Please do fill in `routes` section for automated testing instead of bypassing it. You can access the common parameter using `ctx.query.limit`

seven4x:
> Please do fill in `routes` section for automated testing instead of bypassing it. You can access the common parameter using `ctx.query.limit`

thank you @TonyRL  Before, I didn't realize the significance of the routes section.
and addressed your comments

PR #13696, opened 2023-11-05
fix: Update pixiv's novel original link format

yamada321:

None, or this PR is an issue: the router can fetch novel posts from a pixiv user correctly, but parser gives invalid original urls.  

NOROUTE

NOROUTE

## Note / 说明

The link before this commit itself is a valid format, but only for novel series (as the url implies). In my experience for a couple of weeks, all updates were pushed from individual novel posts, not novel series. Moreover, I believe pixiv's webUI never list series as new items, only individual posts. Therefore I think this change should be fine, but I'm quite new to following pixiv via rsshub.


yamada321:
That was some confusing description about how to use "NOROUTE".

PR #10224, opened 2022-07-15
fix(template):  `routes` example in pr template

superpung:
## 该 PR 相关 Issue / Involved issue

Close #

## 完整路由地址 / Example for the proposed route(s)

```routes
NOROUTE
```

## 新 RSS 检查列表 / New RSS Script Checklist
  
- [ ] 新的路由 New Route
  - [ ] 跟随 [v2 路由规范](https://docs.rsshub.app/joinus/script-standard.html) Follows [v2 Script Standard](https://docs.rsshub.app/en/joinus/script-standard.html)
- [ ] 文档说明 Documentation
  - [ ] 中文文档 CN
  - [ ] 英文文档 EN
- [ ] 全文获取 fulltext
  - [ ] 使用缓存 Use Cache
- [ ] 反爬/频率限制 anti-bot or rate limit?
  - [ ] 如果有, 是否有对应的措施? If yes, do your code reflect this sign?
- [ ] [日期和时间](https://docs.rsshub.app/joinus/pub-date.html) [date and time](https://docs.rsshub.app/en/joinus/pub-date.html)
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] 添加了新的包 New package added
- [ ] `Puppeteer`

PR #14415, opened 2024-02-06
chore: update the "routes" example in the pull request template to make it less confusing

piglei:

Close #

```routes
NOROUTE
```

- [ ] New Route / 新的路由
  - [ ] Follows [v2 Script Standard](https://docs.rsshub.app/joinus/advanced/script-standard) / 跟随 [v2 路由规范](https://docs.rsshub.app/zh/joinus/advanced/script-standard)
- [ ] Documentation / 文档说明
- [ ] Full text / 全文获取
  - [ ] Use cache / 使用缓存
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

In `.github/PULL_REQUEST_TEMPLATE.md`, the code example in the routes section uses the wrong word. If a careless user (like me) copies directly from the example, the PR bot will not be able to pick the route value because the label is incorrect.

This PR fixes this. It also updated `identify.js` so that the "routes" example in the comments won't be consider as a match.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,7 @@ Fail to comply will result in your pull request being closed automatically.
 请在 `routes` 区域填写以 / 开头的完整路由地址，否则你的 PR 将会被无条件关闭。
 如果路由包含在文档中列出可以完全穷举的参数（例如分类），请依次全部列出。
 
-```route
+```routes
 /some/route
 /some/other/route
 /dont/use/this/or/modify/it
--- a/scripts/workflow/test-route/identify.js
+++ b/scripts/workflow/test-route/identify.js
@@ -4,7 +4,9 @@ const allowedUser = new Set(['dependabot[bot]', 'pull[bot]']); // dependabot and
 module.exports = async ({ github, context, core }, body, number, sender) => {
     core.debug(`sender: ${sender}`);
     core.debug(`body: ${body}`);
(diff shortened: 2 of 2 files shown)


TonyRL:
Please take a look at https://github.com/DIYgod/RSSHub/pull/10224#discussion_r922444630

piglei:
> Please take a look at [#10224 (comment)](https://github.com/DIYgod/RSSHub/pull/10224#discussion_r922444630)

I see, so it's not a typo. I will close the current PR.

PS: Maybe we can tweak the bot a bit to fix this. For example, if we remove the comment block (`<!-- ... -->`) before doing the regular expression match in `identify.js`, the correct label "routes" can be used and it won't be considered a match.

piglei:
This change may look like this: 

```
--- a/scripts/workflow/test-route/identify.js
+++ b/scripts/workflow/test-route/identify.js
@@ -4,7 +4,9 @@ const allowedUser = new Set(['dependabot[bot]', 'pull[bot]']); // dependabot and
 module.exports = async ({ github, context, core }, body, number, sender) => {
     core.debug(`sender: ${sender}`);
     core.debug(`body: ${body}`);
-    const m = body.match(/```routes\s+([\S\s]*?)```/);
+    // Remove all HTML comments
+    const bodyNoCmt = body.replaceAll(/<!--[\S\s]*?-->/g, '');
+    const m = bodyNoCmt.match(/```routes\s+([\S\s]*?)```/);
     core.debug(`match: ${m}`);
     let res = null;
```

What do you think? @TonyRL

PR #14668, opened 2024-03-05
feat: opentelementry support

incubator4:

Close #

```routes
NOROUTE
```

- [ ] New Route / 新的路由
- [x] Documentation / 文档说明
- [ ] Full text / 全文获取
  - [ ] Use cache / 使用缓存
  - [ ] Parsed / 可以解析
- [x] New package added / 添加了新的包
- [ ] `Puppeteer`

## Note / 说明
Implement opentelementry feature `metrics` and `trace`

### Metrics
metrics endpoint with `/metrics`
labels:
- `method` http method (example: GET/POST)
- `path` request path
- `status` response return status code
- `unit` only for bucket (example: second/millisecond)

The following metric name supported for now.
- `request_total` record request with  params
- `request_error_total` record request error with params
- `request_duration_seconds_bucket` record request second with params
- `request_duration_milliseconds_bucket` record request millisecond with params

### trace
trace with opentelementry OTLP exporter, which could be collected by jaeger (for example)
endpoint follow [OpenTelemetry Specification on Protocol Exporter.](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options)

### DIFF
--- a/lib/app.ts
+++ b/lib/app.ts
@@ -12,6 +12,7 @@ import debug from '@/middleware/debug';
 import header from '@/middleware/header';
 import antiHotlink from '@/middleware/anti-hotlink';
 import parameter from '@/middleware/parameter';
+import trace from '@/middleware/trace';
 
 import logger from '@/utils/logger';
 
@@ -27,6 +28,7 @@ const app = new Hono();
 app.use(compress());
 
 app.use(mLogger);
+app.use(trace);
 app.use(sentry);
 app.use(accessControl);
 app.use(debug);
--- a/lib/utils/otel/metric.ts
+++ b/lib/utils/otel/metric.ts
@@ -1,0 +1,66 @@
+import { Resource } from '@opentelemetry/resources';
+import { PrometheusExporter, PrometheusSerializer } from '@opentelemetry/exporter-prometheus';
+import { SEMRESATTRS_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
+import { MeterProvider } from '@opentelemetry/sdk-metrics';
+import { Attributes } from '@opentelemetry/api';
+
+interface IMetricAttributes extends Attributes {
+    method: string;
+    path: string;
+    status: number;
+}
+
+interface IHistogramAttributes extends IMetricAttributes {
+    unit: string;
+}
+
+const metric_prefix = 'rsshub';
+
+const exporter = new PrometheusExporter({});
+
+const provider = new MeterProvider({
+    resource: new Resource({
+        [SEMRESATTRS_SERVICE_NAME]: 'rsshub',
+    }),
+    readers: [exporter],
+});
+
+const serializer = new PrometheusSerializer();
+
+const meter = provider.getMeter('rsshub');
+
+const request_total = meter.createCounter<IMetricAttributes>(`${metric_prefix}_request_total`);
+const request_error_total = meter.createCounter<IMetricAttributes>(`${metric_prefix}_request_error_total`);
+const request_duration_seconds_bucket = meter.createHistogram<IHistogramAttributes>(`${metric_prefix}_request_duration_seconds_bucket`, {
+    advice: {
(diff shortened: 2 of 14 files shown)


lib/utils/otel/metric.ts
```diff
@@ -0,0 +1,66 @@
...
+        explicitBucketBoundaries: [0.01, 0.1, 1, 2, 5, 15, 30, 60],
+    },
+});
+const request_duration_milliseconds_bucket = meter.createHistogram<IHistogramAttributes>(`${metric_prefix}_request_duration_milliseconds_bucket`, {
+    advice: {
+        explicitBucketBoundaries: [10, 20, 50, 100, 250, 500, 1000, 5000, 15000],
+    },
+});
+
+export const metric_handler = {
```
TonyRL, line 45:
```suggestion
export const metric = {
```

incubator4: I'm not sure if the rename is necessary, the `metric_handler` here is just a wrapper around the aggregated metrics counter

incubator4: Or any other better var name?

TonyRL: [Script Standard](https://docs.rsshub.app/joinus/advanced/script-standard#naming)

TonyRL (commented):
Please fix your git history. There shouldn't be 2000+ changed files.

package.json
```diff
@@ -51,6 +51,13 @@
   "dependencies": {
     "@hono/node-server": "1.8.2",
     "@notionhq/client": "2.2.14",
+    "@opentelemetry/api": "^1.8.0",
+    "@opentelemetry/exporter-prometheus": "^0.49.1",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.49.1",
+    "@opentelemetry/resources": "^1.22.0",
+    "@opentelemetry/sdk-metrics": "^1.22.0",
+    "@opentelemetry/sdk-trace-base": "^1.22.0",
+    "@opentelemetry/semantic-conventions": "^1.22.0",
```
TonyRL, line 60:
```suggestion
    "@opentelemetry/api": "1.8.0",
    "@opentelemetry/exporter-prometheus": "0.49.1",
    "@opentelemetry/exporter-trace-otlp-http": "0.49.1",
    "@opentelemetry/resources": "1.22.0",
    "@opentelemetry/sdk-metrics": "1.22.0",
    "@opentelemetry/sdk-trace-base": "1.22.0",
    "@opentelemetry/semantic-conventions": "1.22.0",
```
Please pin the dependency version.

website/docs/install/config.md
```diff
@@ -158,6 +158,16 @@ Access code is the md5 generated based on the access key + route, eg:
 
 `SENTRY_ROUTE_TIMEOUT`: Report Sentry if route execution takes more than this milliseconds, default to `3000`
 
+## Opentelementry Config
```
TonyRL, line 161:
```suggestion
## OpenTelemetry Config
```

lib/app.ts
```diff
@@ -27,6 +28,7 @@ const app = new Hono();
 app.use(compress());
 
 app.use(mLogger);
+app.use(trace);
```
TonyRL, line 31: If [`DEBUG_INFO`](https://docs.rsshub.app/install/config#logging-configurations) is set to `false` or any strings other than `true`, the debug info from RSSHub will be hidden and only available if the same string is provided as query string `debug`.

`/metrics` does not respect this behaviour currently.

The same also apply to [`ACCESS_KEY`](https://docs.rsshub.app/install/config#access-keycode).
https://github.com/DIYgod/RSSHub/blob/15a51bdffb7515504e13bb10b36e70be504a8ca7/lib/middleware/access-control.ts#L15

incubator4: Basicly, metrics should be export with a external local port. In that case rsshub may handle another port with `127.0.0.1:9090` for example.
I'm not sure if an external port should be created, but it might be a better solution than modify if else for access-control.
And also should be with env to control local server enable or not.

PR #14909, opened 2024-03-22
feat(route): Add gamer520(全球游戏交流中心)

xzzpig:

Close #

```routes
/gamer520/switchyouxi
/gamer520/3ajuzuo
```

- [x] New Route / 新的路由
  - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明
具有cloudflare保护，需要私人部署才行

### DIFF
--- a/lib/routes/gamer520/index.ts
+++ b/lib/routes/gamer520/index.ts
@@ -1,81 +1,133 @@
 import { Data, DataItem, Route } from '@/types';
 import got from '@/utils/got';
 import { parseDate } from '@/utils/parse-date';
-import { load } from 'cheerio';
 import { Context } from 'hono';
 
 export const route: Route = {
-    path: '/:category?/:order?/:price?',
+    path: '/:category?/:order?',
     categories: ['game'],
     example: '/gamer520/switchyouxi',
     parameters: {
         category: '分类，见下表',
-        order: '排序，见下表',
-        price: '价格, 全部:0;免费1;付费:2;',
+        order: '排序，发布日期: date; 修改日期: modified',
     },
     features: {
-        requireConfig: false,
-        requirePuppeteer: false,
-        antiCrawler: false,
-        supportBT: false,
-        supportPodcast: false,
-        supportScihub: false,
+        antiCrawler: true,
     },
-    name: '全球游戏交流中心',
+    name: '文章',
     maintainers: ['xzzpig'],
     handler,
     url: 'www.gamer520.com/',
     description: `分类
 
-  | Switch游戏下载 | 金手指 | 3A巨作 | switch主题 | PC游戏 | PC游戏下载 | 模拟器大全 | 听我所听 |
-  | --- | --- | --- | --- | --- | --- | --- | --- |
-  | switchyouxi | jinshouzhi | 3ajuzuo | zhuti | pcgame | pcgame | zhangji | ting |
-  
-  排序
-  
-  | 发布日期 | 修改日期 | 评论数量 | 随机 | 热度 |
-  | --- | --- | --- | --- | --- |
-  | date | modified | comment_count | rand | hot |`,
+  | 所有 | Switch游戏下载 | 金手指 | 3A巨作 | switch主题 | PC游戏 |
+  | --- | --- | --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


lib/routes/gamer520/index.ts
```diff
@@ -0,0 +1,81 @@
...
+                title: $titleItem.text().trim(),
+                link: $titleItem.attr('href')!,
+                pubDate: parseDate($item.find('time').attr('datetime')?.trim() ?? ''),
+                category: $item
+                    .find('a[rel="category"]')
+                    .toArray()
+                    .map((a) => $(a).text().trim()),
+                description: $item.html() ?? '',
+            };
+        });
```
TonyRL, line 73: Please use [WordPress REST API](https://developer.wordpress.org/rest-api/) instead.
The API endpoint references are listed at `/wp-json/wp/v2`.

You can list recent posts using `/wp-json/wp/v2/posts`. If you want categories, tags and author for each post, you can add the `_embed` search parameter when calling the posts API.

To adjust the no. of articles returned in the posts API, you can pass the `per_page` search parameter.

---
For categories, you can find the category id first using `/wp-json/wp/v2/categories` with the search parameter `slug`. The id for `switchyouxi` is `2`. Then you can list recent posts with the required tag using /wp-json/wp/v2/posts with the search parameter `categories`, which value will be `2` for `switchyouxi`.

xzzpig: wow, got it. I'll revise it.

xzzpig: A sad story, this site is more anti-bot than I thought. But the strange thing is that when I use the `curl` request, the request succeeds even if the UA is not set. But when I use `got` to request, even if I set up UA, the request will be recognized and blocke ip immediately. I don't know much about `got`, is there any characteristic of `got` that can be accurately recognized? (I tried setting ua and http2 parameters, but it didn't work)
Otherwise, I think `Puppeteer` would be the final solution

lib/routes/gamer520/index.ts
```diff
@@ -0,0 +1,133 @@
...
+    },
+    name: '文章',
+    maintainers: ['xzzpig'],
+    handler,
+    url: 'www.gamer520.com/',
+    description: `分类
+
+  | 所有 | Switch游戏下载 | 金手指 | 3A巨作 | switch主题 | PC游戏 |
+  | --- | --- | --- | --- | --- | --- |
+  | all | switchyouxi | jinshouzhi | 3ajuzuo | zhuti | pcgame |`,
```
TonyRL, line 25:
```suggestion
  | 所有 | Switch 游戏下载 | 金手指     | 3A 巨作 | switch 主题 | PC 游戏 |
  | ---- | --------------- | ---------- | ------- | ----------- | ------- |
  | all  | switchyouxi     | jinshouzhi | 3ajuzuo | zhuti       | pcgame  |`,
```

lib/routes/gamer520/index.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+// (await (await fetch('https://www.gamer520.com/wp-json/wp/v2/categories')).json()).map((category) => ({ slug: category.slug, id: category.id, name: category.name, link: category.link }));
+const CATEGORY_ARRAY = [
```
TonyRL, line 29: > To adjust the no. of articles returned in the posts API, you can pass the `per_page` search parameter.

https://github.com/DIYgod/RSSHub/pull/14909#discussion_r1535799581

FYI, this applies to most of the WordPress API, including `/wp/v2/categories`

lib/routes/gamer520/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    }) as unknown as { data: Post[] };
+
+    const items: DataItem[] = data
+        .map((item) => ({
+            guid: `gamer520:${item.id}`,
+            title: item.title.rendered,
+            link: item.guid.rendered,
+            pubDate: parseDate(item.date_gmt),
+            updated: parseDate(item.modified_gmt),
+            category: item.categories?.map((c) => categories.find((ca) => ca.id === c)?.name ?? '').filter((c) => c !== '') ?? [],
```
xzzpig, line 76: Here we need to map back to `name` based on the `id` of the category, because it is not very numerous and avoids too many requests, so I requested to get all the categories

PR #15275, opened 2024-04-17
feat(route): add route for BBC Twitter accounts

danilo-delbusso:

N/A

```routes
/twitter/bbc-user/BBCBreaking
/twitter/bbc-user/BBCWorld
/twitter/bbc-user/BBCNews
/twitter/bbc-user/BBCNewsAsia
/twitter/bbc-user/BBCAfrica
/twitter/bbc-user/BBCBusiness
/twitter/bbc-user/BBCLondonNews
/twitter/bbc-user/BBCNorthAmerica
/twitter/bbc-user/BBCSport
```

- [x] New Route / 新的路由
  - [x] Parsed / 可以解析
- [ ] `Puppeteer`

### Reason for adding route
BBC Twitter accounts may contain feeds different from those available on bbc.co.uk or similar.

For instance, the Breaking News feed on Twitter contains fewer posts and is more relevant for just major news.

### Based on existing routes
This route uses Twitter Feed fetched from matching BBC accounts, and then uses the existing BBC route to fetch the articles and add their content on the `description`, which avoids having to open the URLs. It also makes use of the `twitter/user` route as a base route to fetch the tweets in the first place.

Note this only works on BBC Twitter accounts that posts BBC Articles, such as twitter.com/BBCBreaking, twitter.com/BBCNews, twitter.com/BBCWorld, etc. More can be seen here: https://twitter.com/search?q=BBC&f=user


danilo-delbusso:
Locally seems to work just fine - I assume all the 503 errors here are related to Twitter API / request limitations? Someone more knowledgeable might know

danilo-delbusso:
keep alive please

danilo-delbusso:
alive

danilo-delbusso:
not stale

danilo-delbusso:
Not stale

danilo-delbusso:
not stale

danilo-delbusso:
not stale

pseudoyu:
We have `/bbc` route and `/twitter` routes, does the `/twitter/bbc-user` have more advantages? seems confusing.

pseudoyu:
> Locally seems to work just fine - I assume all the 503 errors here are related to Twitter API / request limitations? Someone more knowledgeable might know

yes it's due to the Twitter API limit.

danilo-delbusso:
> We have `/bbc` route and `/twitter` routes, does the `/twitter/bbc-user` have more advantages? seems confusing.

good question!

I have been avoiding using `/bbc` because the official BBC RSS feeds include a wide range of news, not just breaking stories, which can add a lot of noise. The Twitter account BBCBreaking is much more focused on major updates, which makes it ideal for a cleaner feed. 

By creating a new feed instead of only using the Twitter one, we get the best of both worlds: we pull in verified BBC articles for reliability, while keeping the feed streamlined and to the point, much like a curated BBC RSS feed with the immediacy of Twitter.

pseudoyu:
> > We have `/bbc` route and `/twitter` routes, does the `/twitter/bbc-user` have more advantages? seems confusing.
> 
> good question!
> 
> I have been avoiding using `/bbc` because the official BBC RSS feeds include a wide range of news, not just breaking stories, which can add a lot of noise. The Twitter account BBCBreaking is much more focused on major updates, which makes it ideal for a cleaner feed.
> 
> By creating a new feed instead of only using the Twitter one, we get the best of both worlds: we pull in verified BBC articles for reliability, while keeping the feed streamlined and to the point, much like a curated BBC RSS feed with the immediacy of Twitter.

got it. But I still think it should belong to `/bbc` namespace instead of twitter. Or it will make users confusing.

danilo-delbusso:
> > > We have `/bbc` route and `/twitter` routes, does the `/twitter/bbc-user` have more advantages? seems confusing.
> > 
> > 
> > good question!
> > I have been avoiding using `/bbc` because the official BBC RSS feeds include a wide range of news, not just breaking stories, which can add a lot of noise. The Twitter account BBCBreaking is much more focused on major updates, which makes it ideal for a cleaner feed.
> > By creating a new feed instead of only using the Twitter one, we get the best of both worlds: we pull in verified BBC articles for reliability, while keeping the feed streamlined and to the point, much like a curated BBC RSS feed with the immediacy of Twitter.
> 
> got it. But I still think it should belong to `/bbc` namespace instead of `/twitter`. Or it will make users confusing.

ah! I had misunderstood your original point. Will make the changes, bear with :)

TonyRL:
It's noticed Twitter implements stricter anti-crawling measures against RSSHub recently.  Fetching data from Twitter, especially content also available on bbc.com, could impact other users' access to information uniquely available on Twitter.  For now, while these restrictions are in place and until Twitter potentially revises its anti-crawling policies or provides more affordable API options, it's suggested to avoid fetching data from Twitter if it's already accessible through bbc.com.

Regarding the `/bbc/*` and `/twitter/*` routes, as well as other similar routes, the intention is to display content directly from the respective source. To maintain consistency and prevent potential issues with Twitter's anti-crawling mechanisms, it's suggested to avoid any "jumping". This prevents RSSHub inadvertently triggering Twitter's restrictions.

PR #15417, opened 2024-04-30
fix(route): fix crossbell character and the following output

markbang:

Close #

```routes
/crossbell/feeds/following/10
/crossbell/notes/character/10
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

### for following
Clip_2024-04-30_19-57-51
Previously, `undefined` would be parsed because `UPDATE_CHARACTER_METADATA`

### for character
Reactivate and serve new RSS `<item>`

### DIFF
--- a/lib/routes/crossbell/notes/character.ts
+++ b/lib/routes/crossbell/notes/character.ts
@@ -1,5 +1,6 @@
 import { Route } from '@/types';
 import got from '@/utils/got';
+import { getItem } from './utils';
 
 export const route: Route = {
     path: '/notes/character/:characterId',
@@ -21,7 +22,7 @@ export const route: Route = {
         },
     ],
     name: 'Notes of character',
-    maintainers: ['DIYgod', 'markbang'],
+    maintainers: ['DIYgod'],
     handler,
     url: 'crossbell.io/*',
 };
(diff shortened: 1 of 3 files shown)


lib/routes/crossbell/notes/character.ts
```diff
@@ -1,6 +1,5 @@
 import { Route } from '@/types';
 import got from '@/utils/got';
-import { getItem } from './utils';
```
TonyRL, line 3: You don't need to change this file at all.
All you need to fix the export of `utils` to something like:
```diff
-export default {
-    getItem: (note) => {
+export const getItem = (note) => {
```

markbang: Thanks for the reply. I think the above format may be better for character RSS output.
can I modify that format to `utils`.

markbang: Or which format is more suitable for character. For example, name should be handle instead of id, so response_getCharacter() is needed.

TonyRL: The current format of `getItem` is already fine.

markbang: Okay, do I need to change character getItem to utils' ?

markbang: > The current format of `getItem` is already fine.

I have modified it as expected. Thanks for the correction. You did a great job😘

PR #15625, opened 2024-05-18
fix(route):fixed the parameter section and add an English site.

Duuckjing:

Close #

```routes
/shisu/news/news
/shisu/news/notice
/shisu/en/features
```

- [x] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

### DIFF
--- a/lib/routes/shisu/en.ts
+++ b/lib/routes/shisu/en.ts
@@ -1,0 +1,61 @@
+import { Route } from '@/types';
+import { load } from 'cheerio';
+import { parseDate } from '@/utils/parse-date';
+import { got } from 'got';
+import cache from '@/utils/cache';
+
+export const route: Route = {
+    path: '/en/features',
+    categories: ['university'],
+    example: '/shisu/en/features',
+    features: {
+        requireConfig: false,
+        requirePuppeteer: false,
+        antiCrawler: false,
+        supportBT: false,
+        supportPodcast: false,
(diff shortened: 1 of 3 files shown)


lib/routes/shisu/en.ts
```diff
@@ -0,0 +1,61 @@
+import { Route } from '@/types';
+import { load } from 'cheerio';
+import { parseDate } from '@/utils/parse-date';
+import { got } from 'got';
```
TonyRL, line 4: `got` is a dev dependency and is not available in production environment.

Duuckjing: There are still many routes using got, so I tried it, but ofetch doesn't seem to support rejectUnauthorized?. Now using puppeteer.

TonyRL: Those routes are using `@/utils/got` which is wrapper of ofetch.

Duuckjing: Yep, and since it doesn't seem to support the insecure rejectUnauthorized: false, I have to use puppeteer instead.

Duuckjing:
Still using ofetch for finally i found it is just using http rather than https, though shown https.

Duuckjing:
I didn't need to open pr again in my repo... and it caused conflicts. The valid one was 0070434.

PR #15857, opened 2024-06-09
ci(test): enable the GitHub Actions reporter for Vitest

FranklinYu:

N/A

```routes
NOROUTE
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

This should publish the JUnit report for human consumption. The raw XML is too long, so not as readable for PR reviewers.

Reference:
- https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,9 +68,14 @@ jobs:
       - name: Build routes
         run: pnpm build
       - name: Test all and generate coverage
-        run: pnpm run vitest:coverage
+        run: pnpm run vitest:coverage --outputFile=tmp/vitest-coverage.xml
         env:
           REDIS_URL: redis://localhost:${{ job.services.redis.ports[6379] }}/
+      - name: Publish test report
+        uses: mikepenz/action-junit-report@v4
+        if: ${{ matrix.node-version == '20' }}
+        with:
+          report_paths: 'tmp/vitest-coverage.xml'
       - name: Upload coverage to Codecov
         if: ${{ matrix.node-version == '20' }}
         uses: codecov/codecov-action@v4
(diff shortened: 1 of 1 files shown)


FranklinYu:
Note: `include_passed` is false by default. Let me know if you prefer that turned on (so it will post the report even if everything passes).

.github/workflows/test.yml
```diff
@@ -68,9 +69,14 @@ jobs:
       - name: Build routes
         run: pnpm build
       - name: Test all and generate coverage
-        run: pnpm run vitest:coverage
+        run: pnpm run vitest:coverage --outputFile=tmp/vitest-coverage.xml
```
TonyRL, line 72: This breaks 
https://github.com/DIYgod/RSSHub/blob/044d17fb3366ed9e01e244d1b0a175987c08587b/.github/workflows/test.yml#L80-L84
since it can no longer read the HTML generated from coverage folder.

FranklinYu: What do you mean by “HTML generated from coverage folder”? When I run the `vitest:coverage` script (without the flag), it doesn’t generate any HTML in that folder either, only a bunch of JSON files like `coverage/.tmp/coverage-x.json`. Were you referring to those JSON files?

TonyRL: Please run the test with `GITHUB_ACTIONS=1`.

FranklinYu: I searched `GITHUB_ACTIONS` in Vitest documentation, and found [GitHub action reporter](https://vitest.dev/guide/reporters.html#github-actions-reporter). I think it is a better solution than the solution I proposed, for two reasons:

1. We don’t need to introduce a new dependency.
2. Less likely to break the Codecov job.

I updated the PR title to reflect that. I assumed that we squash commits when merging, so I pushed a new commit instead of amending the existing one.

PR #16246, opened 2024-07-24
Restore podcast enclosures in kemono feeds.

tinruufu:

```routes
/kemono/patreon/5437626
/kemono/patreon/5461428
/kemono/patreon/44261000
```

## Note / 说明

this restores a feature originally introduced in #13821 that was removed in b2f3bcc5c36a8cde7748de09ddcac3251400699f, allowing these feeds to be used in a podcast client.

### DIFF
--- a/lib/routes/kemono/index.ts
+++ b/lib/routes/kemono/index.ts
@@ -140,6 +140,26 @@ async function handler(ctx) {
                     desc += kemonoFile;
                 }
 
+                let enclosureInfo = {};
+                load(desc)('audio source, video source').each(function () {
+                    const src = $(this).attr('src') ?? '';
+                    const mimeType =
+                        {
+                            m4a: 'audio/mp4',
+                            mp3: 'audio/mpeg',
+                            mp4: 'video/mp4',
+                        }[src.replace(/.*\./, '').toLowerCase()] || null;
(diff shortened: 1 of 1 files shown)


tinruufu:
the `src` on these enclosures isn't being given the right domain. im not sure how it's being added in the description; i thought i was pulling it out verbatim

lib/routes/kemono/index.ts
```diff
@@ -140,13 +140,34 @@ async function handler(ctx) {
                     desc += kemonoFile;
                 }
 
+                let enclosureInfo = {};
+                load(desc)('audio source, video source').each(function () {
```
TonyRL, line 144:
```suggestion
                $('audio source, video source').each(function () {
```
reuse
https://github.com/DIYgod/RSSHub/blob/65865f6eee1a098b2236dd95cfbac36febf065de/lib/routes/kemono/index.ts#L128

tinruufu: i tried that at first. it doesn't work, because at the time `$` was last loaded, `desc` did not contain the media elements i'm looking for

i guess i could do another round of `$ = load(desc)`, but im only using it once so it feels unnecessary to assign it

PR #16464, opened 2024-08-17
feat(middleware): process feed title and description using LLM

coxde:

Close #

```routes
NOROUTE
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明
Add a title translation feature using LLM, I'll open a new PR later for the [docs](https://github.com/RSSNext/rsshub-docs).


TonyRL (commented):
This can be done by modifying https://github.com/DIYgod/RSSHub/blob/c2937395f1e1e2bd70a6b649cfcc94cb4d4096ff/lib/middleware/parameter.ts#L348-L372
to pass the title as well instead of adding a new common parameter

coxde:
> This can be done by modifying
> 
> https://github.com/DIYgod/RSSHub/blob/c2937395f1e1e2bd70a6b649cfcc94cb4d4096ff/lib/middleware/parameter.ts#L348-L372
> 
> to pass the title as well instead of adding a new common parameter

Indeed. Maybe I could also provide an option for passing all, title only or description only? For my case, I just need to process the title.

TonyRL:
You may add an extra env like `config.openai.inputOption` which takes either `title`, `description` or `both` with `description` being the default one so it won't break the current flow.

coxde:
> You may add an extra env like `config.openai.inputOption` which takes either `title`, `description` or `both` with `description` being the default one so it won't break the current flow.

I've done the refactoring, feel free to check out any further improvements as my knowledge of JavaScript is limited.

Besides `OPENAI_INPUT_OPTION`, I also separated the prompts for title and description, which might be useful for `both` mode. I kept the original prompt for compatibility. But this is an example that works best in my case.
```
OPENAI_PROMPT_TITLE='Please translate the following title into Simplified Chinese. Output ONLY translated text.'
OPENAI_PROMPT=Please summarize the following article CONCISELY into Simplified Chinese and reply with markdown format. Output ONLY summarized text.
```

lib/middleware/parameter.ts
```diff
@@ -327,23 +327,53 @@ const middleware: MiddlewareHandler = async (ctx, next) => {
...
-                                }
-                                const summary_md = await summarizeArticle(text);
-                                return md.render(summary_md);
+                    try {
+                        // handle description
+                        if (config.openai.inputOption === 'description' && item.description) {
+                            const description = await cache.tryGet(`openai:description:${item.link}`, async () => {
+                                const description = convert(item.description!);
+                                const description_md = await getAiCompletion(config.openai.promptDescription, description);
+                                return md.render(description_md);
```
TonyRL, line 336: `camelCase`

lib/middleware/parameter.ts
```diff
@@ -327,23 +327,53 @@ const middleware: MiddlewareHandler = async (ctx, next) => {
...
+                            });
+                            // replace the title
+                            if (title !== '') {
+                                item.title = title + '';
+                            }
+
+                            const description = await cache.tryGet(`openai:description:${item.link}`, async () => {
+                                const description = convert(item.description!);
+                                const description_md = await getAiCompletion(config.openai.promptDescription, description);
+                                return md.render(description_md);
```
TonyRL, line 368: `camelCase`

TonyRL (commented):
Please update the test case as well

coxde:
> Please update the test case as well

Could you kindly take a look? Not sure why there are two coverage missing.

PR #16468, opened 2024-08-17
feat(towardsdatascience): add route

mintyfrankie:

Close #

```routes
/towardsdatascience/latest
```

- [X] New Route / 新的路由
  - [X] Follows [Script Standard](https://docs.rsshub.app/joinus/advanced/script-standard) / 跟随 [路由规范](https://docs.rsshub.app/zh/joinus/advanced/script-standard)
- [X] [Date and time](https://docs.rsshub.app/joinus/advanced/pub-date) / [日期和时间](https://docs.rsshub.app/zh/joinus/advanced/pub-date)
  - [X] Parsed / 可以解析
  - [X] Correct time zone / 时区正确
- [ ] `Puppeteer`

## Note / 说明

Use [freedium](https://freedium.cfd) as proxy to fetch full-text contents.

### DIFF
--- a/lib/routes/towardsdatascience/latest.ts
+++ b/lib/routes/towardsdatascience/latest.ts
@@ -31,32 +31,25 @@ async function handler() {
     });
     const data = JSON.parse(response.slice(16));
 
-    const list = $('div.postArticle')
-        .toArray()
-        .map((item) => {
-            item = $(item);
-
-            const a = item.find('div.postArticle-content a');
-            const title = a.find('h3').text();
-            const originalLink = a.attr('href').split('?')[0];
-            const link = `https://freedium.cfd/${originalLink}`;
-
-            const header = item.find('div.postMetaInline-authorLockup');
-            const author = header.find('a.ds-link').text().replace('Towards Data Science', '').trim();
-            const pubDate = parseDate(header.find('time').attr('datetime'));
-
-            return {
-                title,
-                link,
-                author,
-                pubDate,
-            };
-        });
+    const list = data.payload.posts.map((item) => {
+        const title = item.title;
+        const link = `https://towardsdatascience.com/${item.uniqueSlug}`;
+        const freediumLink = `https://freedium.cfd/https://towardsdatascience.com/${item.uniqueSlug}`;
+        const author = data.payload.references.User[item.creatorId].name;
+        const pubDate = parseDate(item.createdAt);
+        return {
+            title,
+            link,
+            freediumLink,
+            author,
+            pubDate,
+        };
+    });
 
     const items = await Promise.all(
         list.map((item) =>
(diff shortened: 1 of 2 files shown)


lib/routes/towardsdatascience/latest.ts
```diff
@@ -0,0 +1,70 @@
...
+    handler,
+};
+
+async function handler() {
+    const baseUrl = 'https://towardsdatascience.com/latest';
+    const feedLang = 'en';
+    const feedDescription = 'Latest articles from Towards Data Science';
+
+    const response = await ofetch(baseUrl);
+    const $ = load(response);
```
TonyRL, line 28:
```suggestion
    const response = await ofetch('https://medium.com/towards-data-science/latest?posts=true', {
        headers: {
            accept: 'application/json',
        },
    });
    const data = JSON.parse(response.slice(16));
```

lib/routes/towardsdatascience/latest.ts
```diff
@@ -0,0 +1,70 @@
...
+            const author = header.find('a.ds-link').text().replace('Towards Data Science', '').trim();
+            const pubDate = parseDate(header.find('time').attr('datetime'));
+
+            return {
+                title,
+                link,
+                author,
+                pubDate,
+            };
+        });
```
TonyRL, line 50:
```suggestion
    const list = data.payload.posts.map((item) => {
        const title = item.title;
        const link = `https://towardsdatascience.com/${item.uniqueSlug}`;
        const freediumLink = `https://freedium.cfd/https://towardsdatascience.com/${item.uniqueSlug}`;
        const author = data.payload.references.User[item.creatorId].name;
        const pubDate = parseDate(item.createdAt);

        return {
            title,
            link,
            freediumLink,
            author,
            pubDate,
        };
    });
```
I'm not sure how freedium works. You may need to update `freediumLink`

mintyfrankie: Thanks Tony! Always wondering how you found out these API calls; would you mind sharing your method? I depends on console and also package capture tool, but not too familiar with them.

Regarding freedium, your approach is correct. But somehow I am not sure if the field `freediumLink` will be recognized by RSS readers. Users can open links to full-text content (behind paywall) directly if we put freedium link as `link` itself. What do you think?

mintyfrankie: Btw, if we assign `freediumLink` here, we will also have to change the object got by `cache.tryGet`, because we won't get content with the original links without a subscription, yet with freedium it's very probable.

mintyfrankie: Thanks for the off-topic! Definitely learnt something.

mintyfrankie:
Should be okay now.

PR #16527, opened 2024-08-24
feat(route): add goldman sachs developer blog

chesha1:

Close #

```routes
/gs/developer/blog
```

- [x] New Route / 新的路由
  - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明
It seems that server side only checks user-agent to prevent spiders, so I use a fake browser-like user-agent.

### DIFF
--- a/lib/routes/gs-developer/blog.ts
+++ b/lib/routes/gs-developer/blog.ts
@@ -1,0 +1,64 @@
+import { Route, Data } from '@/types';
+import ofetch from '@/utils/ofetch';
+import { load } from 'cheerio';
+
+export const route: Route = {
+    path: '/blog',
+    categories: ['blog'],
+    example: '/gs-developer/blog',
+    parameters: {},
+    features: {
+        requireConfig: false,
+        requirePuppeteer: false,
+        antiCrawler: false,
+        supportBT: false,
+        supportPodcast: false,
+        supportScihub: false,
+    },
+    radar: [
+        {
+            source: ['developer.gs.com/blog/posts'],
+            target: '/blog',
+        },
+    ],
+    name: 'blog',
+    zh: {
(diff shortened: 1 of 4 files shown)


lib/routes/gs-developer/blog.ts
```diff
@@ -0,0 +1,64 @@
+import { Route, Data } from '@/types';
```
TonyRL, line 1: Please use the [namespace](https://docs.rsshub.app/joinus/new-rss/start-code#creating-namespace) `gs` instead of `gs-developer`.

lib/routes/gs-developer/blog.ts
```diff
@@ -0,0 +1,64 @@
...
+    maintainers: ['chesha1'],
+    handler: handlerRoute,
+};
+
+async function handlerRoute(): Promise<Data> {
+    const response = await ofetch('https://developer.gs.com/blog/posts', {
+        headers: {
+            'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36',
+        },
+        body: null,
```
TonyRL, line 37: RSSHub do use a randomised UA
https://github.com/DIYgod/RSSHub/blob/69cbeae1826b1107523903beebe8cbb9335917ad/lib/config.ts#L380
Are there any specific reasons to use this fixed UA string instead of a randomised one?

chesha1: No other reasons, just because I do not know rsshub will have randomised UAs.

And this does not work in my local environment.  Does this randomised UA only work in rsshub server side?

PR #15388, opened 2024-04-27
fix(route): 修复 hellogithub 路由错误

CaoMeiYouRen:
## Involved Issue / 该 PR 相关 Issue

Close #

## Example for the Proposed Route(s) / 路由地址示例

```routes
/hellogithub/volume
/hellogithub/month
/hellogithub/ranking
/hellogithub/report
/hellogithub/article
```

## New RSS Route Checklist / 新 RSS 路由检查表
  
- [ ] New Route / 新的路由
  - [ ] Follows [Script Standard](https://docs.rsshub.app/joinus/advanced/script-standard) / 跟随 [路由规范](https://docs.rsshub.app/zh/joinus/advanced/script-standard)
- [ ] Anti-bot or rate limit / 反爬/频率限制
  - [ ] If yes, do your code reflect this sign? / 如果有, 是否有对应的措施?
- [ ] [Date and time](https://docs.rsshub.app/joinus/advanced/pub-date) / [日期和时间](https://docs.rsshub.app/zh/joinus/advanced/pub-date)
  - [ ] Parsed / 可以解析
  - [ ] Correct time zone / 时区正确
- [ ] New package added / 添加了新的包
- [ ] `Puppeteer`

PR #8879, opened 2022-01-15
feat(route): 增加HelloGithub的RSS订阅

moke8:
## 该 PR 相关 Issue / Involved issue
Hello Github RSS订阅提交
Close #

## 完整路由地址 / Example for the proposed route(s)

```routes
/hellogithub/article        // 文章列表
/hellogithub/ranking/:type      // 编程语言排行榜  type = tiobe 编程语言排行 ,db 数据库排行,webserver 服务端语言排行
/hellogithub/month            // 月刊
```

## 新RSS检查列表 / New RSS Script Checklist
  
- [√ ] New Route
- [ ] Documentation
  - [ ] CN
  - [ ] EN
- [ √ ] 全文获取 fulltext
  - [ ] Use Cache
- [ ] 反爬/频率限制 anti-bot or rate limit?
  - [ ] 如果有, 是否有对应的措施? If yes, do your code reflect this sign?
- [ ] 日期和时间 date and time
  - [ ] 可以解析 Parsed
  - [ ] 时区调整 Correct TimeZone
- [ ] 添加了新的包 New package added 
- [ ] `Puppeteer`

## 说明 / Note
HelloGithub       https://hellogithub.com/
编程语言排行榜、月刊、文章列表等

PR #16823, opened 2024-09-19
fix: update route HelloGitHub

521xueweihan:

Close #

```routes
/hellogithub/volume
/hellogithub/month
/hellogithub/ranking
/hellogithub/report
/hellogithub/home
```

## Note / 说明

接口变更

### DIFF
--- a/lib/routes/hellogithub/index.ts
+++ b/lib/routes/hellogithub/index.ts
@@ -1,24 +1,19 @@
 import { Route } from '@/types';
-import { getCurrentPath } from '@/utils/helpers';
-const __dirname = getCurrentPath(import.meta.url);
-
-import cache from '@/utils/cache';
 import got from '@/utils/got';
+
 import { load } from 'cheerio';
 import { parseDate } from '@/utils/parse-date';
-import { art } from '@/utils/render';
-import path from 'node:path';
 
 const sorts = {
-    hot: '热门',
-    last: '最近',
+    featured: '精选',
+    all: '全部',
 };
 
 export const route: Route = {
-    path: ['/article/:sort?/:id?'],
+    path: ['/home/:sort?/:id?'],
     categories: ['programming'],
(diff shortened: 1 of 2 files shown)


lib/routes/hellogithub/index.ts
```diff
@@ -62,66 +57,21 @@ async function handler(ctx) {
...
         author: item.author,
         link: `${rootUrl}/repository/${item.item_id}`,
-        description: item.description,
         pubDate: parseDate(item.updated_at),
+        name: `${item.author}/${item.name}`,
+        summary: item.summary,
+        language: item.primary_lang,
     }));
 
-    items = await Promise.all(
```
TonyRL, line 86: This essentially remove all the `description` of `/hellogithub/home`
https://github.com/DIYgod/RSSHub/pull/16823#issuecomment-2361584669

521xueweihan: I think /home is a section introduction, not a traversal item /detail

TonyRL: This route does provide these [since the very beginning](https://github.com/DIYgod/RSSHub/blob/e4c6dabf7061fa99ce7adab0847314460bca1a1d/lib/v2/hellogithub/article.js).

521xueweihan: But the ones you mentioned haven't been updated for a long time, right?

TonyRL: The one I mentioned do receive reasonable updates. This route is originally purposed in #8879, reworked in #11000 and received fixes in #15388. The last fix was in Apr this year. All of these can be seen from git commit history, and none of these changes tried to remove the `desription`.

521xueweihan: The point is that these commits did not solve the problem, and the content remained in June. This is the main purpose of my fix

PR #17105, opened 2024-10-13
feat(route): add icable route

quiniapiezoelectricity:

Close #

```routes
/i-cable/news
/i-cable/news/新聞資訊
/i-cable/news/港聞
/i-cable/news/財經資訊
/i-cable/news/兩岸國際
/i-cable/news/中國在線
```

- [x] New Route / 新的路由
  - [x] Parsed / 可以解析
- [ ] `Puppeteer`


lib/routes/icable/news.ts
```diff
@@ -0,0 +1,243 @@
...
+                    method: 'get',
+                    url: item.link,
+                });
+                const $ = load(detailResponse.data);
+                item.article = $('article.post-detail-wrap > p')
+                    .map((i, el) => $(el).html())
+                    .toArray();
+                item.video = $('article.post-detail-wrap video')
+                    .map((i, el) => ({
+                        sourceUrl: $(el).attr('data-url'),
```
TonyRL, line 215: Maybe I've missed something here, but I don't think placing m3u8 link in source will make it playable without aids from external JS lib.

lib/routes/icable/news.ts
```diff
@@ -0,0 +1,243 @@
...
+        },
+        operationName: 'getListsByCate',
+    };
+}
+
+async function handler(ctx) {
+    const category = ctx.req.param('category') ?? '新聞資訊';
+    const limit = ctx.req.query('limit') ?? 20;
+    const embed = ctx.req.query('embed') === '1';
+    const graphqlUrl = 'https://www.i-cable.com/graphql';
```
TonyRL, line 137: Since the site built atop WordPress. You can use the [WordPress REST API](https://developer.wordpress.org/rest-api/) instead.

For categories, you can find the category id first using `/wp-json/wp/v2/categories` with the search parameter `slug`. The id for `新聞資訊` is `37`. Then you can list recent posts with the required categories using `/wp-json/wp/v2/posts` with the search parameter `categories`, which will be `37` for `新聞資訊`. If you want plain categories, tags, post header image for each post in the posts API, you can add the `_embed` search parameter to `/wp-json/wp/v2/posts`

To adjust the no. of articles returned to the posts API, you can pass the `per_page` search parameter to `/wp-json/wp/v2/posts`.

This way, you can finish RSS generation in only 2 requests.

quiniapiezoelectricity:
Thanks for the review @TonyRL 

You are right, the WordPress Rest API is much better than the API I was using, I have rewritten the entire route to use the WordPress API instead.

The video/embed function had been removed since the API doesn't provide the link to the video and it would require fetching every single page to find the source, in addition to the fact that the m3u8 link might not be playable in every RSS reader (?) (ps: it works in my RSS reader of choice, but mileage may vary depends on RSS reader?) 

The category parameter now only supports category name in order for it to utilise the slug search parameter on the API .

Route is also renamed to i-cable

I believe everything is now fixed. Thanks again.

lib/routes/i-cable/news.ts
```diff
@@ -0,0 +1,73 @@
...
+:::`,
+};
+
+async function handler(ctx) {
+    const category = ctx.req.param('category') ?? '新聞資訊';
+    const limit = ctx.req.query('limit') ?? 20;
+    const root = 'https://www.i-cable.com/wp-json/wp/v2';
+
+    const response = await cache.tryGet(`${root}/categories?slug=${category}`, async () => await got(`${root}/categories?slug=${category}`), config.cache.routeExpire, false);
+    const metadata = response.data[0];
```
TonyRL, line 50: `.data[0]` maybe empty if user provides an invalid input so please throw an exception if that happens

quiniapiezoelectricity:
@TonyRL Issue resolved. Thanks

PR #17232, opened 2024-10-21
fix(route/twitter): Prevent storing empty cookie if obtaining such from `auth_token` failed.

dzx-dzx:

Close #

```routes
/twitter/user/DIYgod
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

### DIFF
--- a/lib/routes/twitter/api/web-api/utils.ts
+++ b/lib/routes/twitter/api/web-api/utils.ts
@@ -13,36 +13,40 @@ import login from './login';
 
 let authTokenIndex = 0;
 
-const token2Cookie = (token) =>
-    cache.tryGet(`twitter:cookie:${token}`, async () => {
-        const jar = new CookieJar();
-        jar.setCookieSync(`auth_token=${token}`, 'https://x.com');
-        try {
-            const agent = proxy.proxyUri
-                ? new ProxyAgent({
-                      factory: (origin, opts) => new CookieClient(origin as string, { ...opts, cookies: { jar } }),
-                      uri: proxy.proxyUri,
-                  })
-                : new CookieAgent({ cookies: { jar } });
-            if (token) {
-                await ofetch('https://x.com', {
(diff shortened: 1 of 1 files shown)


AiraNadih:
This commit broke the RSSHub instances that authenticate using `auth_token`.

I suspect the issue might be here:

https://github.com/DIYgod/RSSHub/blob/c8496f370b0ddde572420620332511d0b848e726/lib/routes/twitter/api/web-api/utils.ts#L43

dzx-dzx:
@AiraNadih  It works on my side. What error messages are given?

AiraNadih:
> @AiraNadih It works on my side. What error messages are given?

```
Error in /twitter/home/onlyMedia=1: ConfigNotFoundError: Twitter cookie for token *** is not valid
```

dzx-dzx:
I don't think the token method is stable, even before this pr. Please try a few times more.

AiraNadih:
> I don't think the token method is stable, even before this pr. Please try a few times more.

But after I rolled back to commit 682c943, it was normal.

AiraNadih:
I just tested it again.
Including the two times I tested it when I commented before, that's a total of three times now.

The first request after initial deployment is normal, but subsequent requests will result in errors.

Rolling back to the previous commit works correctly.
Therefore, I am certain that some part of this commit must have broken the relevant logic.

flameleaf:
One strange quirk I've noticed about the auth_token cookie is that quoting the variable sometimes fixes it. It'll randomly fail on my deployment, but if I change it from TWITTER_AUTH_TOKEN=token_name to TWITTER_AUTH_TOKEN='token_name', or vice-versa, it'll return to a usable state.

AiraNadih:
Seeing the latest merged PR #17429, I finally understand where the problem lies.

PR #17310, opened 2024-10-25
feat(twitter): only display tweets that contains a media

Rakambda:

Close #

```routes
/twitter/home
/twitter/likes/DIYgod
/twitter/list/221172095
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明
Add an option to only display tweets that contains a media. Effectively this excludes "text only" tweets, but as the filtering itself is mroe "keep only medias" I prefered to name it that way. I don't know your preferences about it.

Didn't add this option to the `user` route as there is the `media` route that basically does the same.
Added info in the params table, though as it isn't added to every route I'm not too sure (though same params in that list are not on all routes anyways).

### DIFF
--- a/lib/routes/twitter/home-latest.ts
+++ b/lib/routes/twitter/home-latest.ts
@@ -40,7 +40,7 @@ export const route: Route = {
 
 async function handler(ctx) {
     // For compatibility
-    const { count, include_rts, only_medias } = utils.parseRouteParams(ctx.req.param('routeParams'));
+    const { count, include_rts, only_media } = utils.parseRouteParams(ctx.req.param('routeParams'));
     const params = count ? { count } : {};
 
     await api.init();
@@ -48,7 +48,7 @@ async function handler(ctx) {
     if (!include_rts) {
         data = utils.excludeRetweet(data);
     }
-    if (only_medias) {
+    if (only_media) {
         data = utils.keepOnlyMedia(data);
     }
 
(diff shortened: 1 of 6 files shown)


lib/routes/twitter/utils.ts
```diff
@@ -495,4 +496,15 @@ export const excludeRetweet = function (tweets) {
...
-export default { ProcessFeed, getAppClient, parseRouteParams, excludeRetweet };
+export const keepOnlyMedia = function (tweets) {
+    const excluded = [];
+    for (const t of tweets) {
+        if (!t.extended_entities || !t.extended_entities.media) {
+            continue;
+        }
+        excluded.push(t);
+    }
+    return excluded;
```
TonyRL, line 507: This can be simplified into
```suggestion
    const excluded = tweets.filter((t) => t.extended_entities && t.extended_entities.media);
    return excluded;
```

lib/routes/twitter/home-latest.ts
```diff
@@ -40,14 +40,17 @@ export const route: Route = {
 
 async function handler(ctx) {
     // For compatibility
-    const { count, include_rts } = utils.parseRouteParams(ctx.req.param('routeParams'));
+    const { count, include_rts, only_medias } = utils.parseRouteParams(ctx.req.param('routeParams'));
```
TonyRL, line 43: `only_media`
media is uncountable

PR #17941, opened 2024-12-19
feat(route): Add New Route for UPS Tracking

Aquabet:

Close #

```routes
/ups/track/1Z78R6790470567520
```

- [x] New Route / 新的路由
  - [x] Parsed / 可以解析
- [x] `Puppeteer`

## Note / 说明

发现关于`/utils/parse-date`的不明bug，会导致`parseDate('12/17/2024 2:29 PM', 'MM/DD/YYYY h:mm A')`中的`AM/PM`无法正确解析。
UPS的时间记录根据事件发生城市所在时区。数据量太大，短时间无法完成时区正确。


lib/routes/ups/track.ts
```diff
@@ -0,0 +1,107 @@
...
+        supportPodcast: false,
+        supportScihub: false,
+    },
+    radar: [
+        {
+            source: ['ups.com/track?loc=en_US&tracknum=:trackingNumber'],
+            target: '/ups/track/:trackingNumber',
+        },
+    ],
+    name: 'UPS Tracking',
```
TonyRL, line 25: Do not repeat namespace name https://github.com/DIYgod/RSSHub/blob/612d9cd34b99ceb42759180c6e91657af34fa843/lib/routes/ups/namespace.ts#L4 in route name

lib/routes/ups/track.ts
```diff
@@ -0,0 +1,107 @@
...
+    const url = `https://www.ups.com/track?loc=en_US&tracknum=${trackingNumber}`;
+
+    const browser = await puppeteer();
+    const page = await browser.newPage();
+
+    await page.setRequestInterception(true);
+
+    // skip loading images, stylesheets, and fonts
+    page.on('request', (request) => {
+        if (['image', 'stylesheet', 'font'].includes(request.resourceType())) {
```
TonyRL, line 41:
```suggestion
        if (['image', 'stylesheet', 'font', 'ping', 'fetch'].includes(request.resourceType())) {
```

lib/routes/ups/track.ts
```diff
@@ -0,0 +1,107 @@
...
+            request.abort();
+        } else {
+            request.continue();
+        }
+    });
+
+    await page.goto(url, { waitUntil: 'domcontentloaded' });
+
+    await page.waitForSelector('tr[id^="stApp_ShpmtProg_LVP_progress_row_"]', {
+        timeout: 30000,
```
TonyRL, line 51: Do not wait for 30s, the browser will be closed by RSSHub https://github.com/DIYgod/RSSHub/blob/f5fc827bb8bda196844330f5907422c5d8d680ec/lib/utils/puppeteer.ts#L57-L59

lib/routes/ups/track.ts
```diff
@@ -0,0 +1,107 @@
...
+            return {
+                title: `${status}: ${location}`,
+                link: url,
+                description: `
+                    Status: ${status} <br>
+                    Location: ${location} <br>
+                    Date and Time: ${date} ${time}
+                `.trim(),
+                pubDate: parseDate(formattedDateTime, 'MM/DD/YYYY h:mm A'),
+            };
```
TonyRL, line 98: Please create a guid as `link` is not unique

lib/routes/ups/track.ts
```diff
@@ -0,0 +1,107 @@
...
+    await page.waitForSelector('tr[id^="stApp_ShpmtProg_LVP_progress_row_"]', {
+        timeout: 30000,
+    });
+
+    const content = await page.content();
+    browser.close();
+    const $ = load(content);
+
+    // Extract tracking events
+    const items = $('tr[id^="stApp_ShpmtProg_LVP_progress_row_"]')
```
TonyRL, line 59: It would be better if you can intercept the response of `/track/api/Track/GetStatus` since it provides 22 shipment progress activities instead of 6 only (from `/ups/track/1Z78R6790470567520`)

lib/routes/ups/track.ts
Aquabet: UPS does not provide a separate page for each tracking record. That's why I'm linking each record to the tracking page.

lib/routes/ups/track.ts
Aquabet: Using the official UPS API `/track/api/Track/GetStatus` requires OAuth 2.0, which involves account registration and a payment method, so I had to rely on scraping to fetch updates.

lib/routes/ups/track.ts
TonyRL: Without provide a guid for feed item, RSSHub will use `link` as guid. Since `guid` is the same across all items, some RSS readers will show only one entry.

lib/routes/ups/track.ts
TonyRL: Double-check your devtools network logs, this request does show up when you visit the site and I believe it doesn't require OAuth

lib/routes/ups/track.ts
```diff
@@ -0,0 +1,112 @@
...
+        items.push({
+            title: status,
+            link: url,
+            description: `
+                Status: ${status} <br>
+                Location: ${location} <br>
+                Date and Time: ${dateTimeStr}
+            `,
+            pubDate,
+        });
```
TonyRL, line 104: Please use [Array.prototype.map()](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/map) instead.

lib/routes/ups/track.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    await page.waitForSelector('tr[id^="stApp_activitydetails_row"]');
+
+    const content = await page.content();
+    await browser.close();
+
+    const $ = load(content);
+
+    const rows = $('tr[id^="stApp_activitydetails_row"]');
+    const items: { title: string; link: string; description: string; pubDate: Date }[] = [];
```
TonyRL, line 64: Please use https://github.com/DIYgod/RSSHub/blob/0a93a7486da3188e712d6446e59911077637e132/lib/types.ts#L32 instead

PR #18702, opened 2025-03-26
fix(route/iplaysoft): Update to use URL scraping due to API deprecation

LokHsu:

Close #

```routes
/iplaysoft
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

The original scraping method for iplaysoft is no longer working due to the official API ban. Therefore, we now directly scrape new content from iplaysoft via the URL.

### DIFF
--- a/lib/routes/iplaysoft/index.ts
+++ b/lib/routes/iplaysoft/index.ts
@@ -1,17 +1,49 @@
 import { Data, DataItem, Route, ViewType } from '@/types';
-import { fetchNewsItems } from './utils';
+import cache from '@/utils/cache';
+import ofetch from '@/utils/ofetch';
+import parser from '@/utils/rss-parser';
+import { parseDate } from '@/utils/parse-date';
+import { load } from 'cheerio';  // html parser
 
-export const handler = async (): Promise<Data> => {
-    const rootUrl = 'https://www.iplaysoft.com/';
-    const postApiUrl = `${rootUrl}wp-json/wp/v2/posts?_embed`;
+export const handler = async (ctx): Promise<Data> => {
+    const feed = await parser.parseURL('https://feed.iplaysoft.com');
+    const limit = Number.parseInt(ctx.req.query('limit') || '20', 10);
 
-    const items: DataItem[] = await fetchNewsItems(postApiUrl);
+    const filteredItems = feed.items
+        .filter((item) => {
+            if (!item?.link || !item?.pubDate) {
+                return false;
+            }
+            return new URL(item.link).hostname.match(/.*\.iplaysoft\.com$/);
+        })
+        .slice(0, limit) as DataItem[];
+
+    const items: DataItem[] = await Promise.all(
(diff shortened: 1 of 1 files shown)


lib/routes/iplaysoft/index.ts
```diff
@@ -1,31 +1,59 @@
...
 export const route: Route = {
-    path: '/',
-    name: '首页',
-    url: 'www.iplaysoft.com',
-    maintainers: ['williamgateszhao', 'cscnk52'],
-    handler,
-    example: '/iplaysoft',
-    parameters: undefined,
-    description: undefined,
+    path: '/index',
```
TonyRL, line 52: Changing the route path requires all previous subscribers to update their subscription URLs. Is this change necessary?

lib/routes/iplaysoft/index.ts
```diff
@@ -1,31 +1,59 @@
...
     categories: ['program-update'],
+    view: ViewType.Articles,
+    example: '/iplaysoft/index',
+    parameters: {},
```
TonyRL, line 56: Unnecessary reordering

lib/routes/iplaysoft/index.ts
```diff
@@ -35,11 +63,14 @@ export const route: Route = {
         supportPodcast: false,
         supportScihub: false,
     },
+    name: '首页',
+    url: 'www.iplaysoft.com',
```
TonyRL, line 67: Unnecessary reordering

lib/routes/iplaysoft/index.ts
```diff
@@ -35,11 +63,14 @@ export const route: Route = {
...
+    maintainers: ['williamgateszhao', 'cscnk52', 'LokHsu'],
+    handler,
```
TonyRL, line 69: Unnecessary reordering

lib/routes/iplaysoft/index.ts
LokHsu: I think it's not necessary, but I found an issue in the following:

The original route path is '/', which points to the '/iplaysoft/index' file. If we do not replace the path '/' into '/index'. The [github-actions](https://github.com/apps/github-actions) will test the path '/iplaysoft/index' and callback 404 not found. So, we finally have to change the route path in this submission. If the issue doesn't matter, we can resubmit this file to restore the original route path.

Thank you for your review.

lib/routes/iplaysoft/index.ts
LokHsu: Thanks for your review. I just want to reorder the array to improve the relevance of its elements by grouping related items together. If necessary, we can restore the original order. The reviews below are also the same.

lib/routes/iplaysoft/index.ts
TonyRL: GitHub actions will test against whatever is filled in the PR form and you can change the content using the edit button.

PR #19406, opened 2025-06-18
feat(route/javdb): support category exclusion

sgpublic:

Close #

```routes
/javdb/actors/R2Vg/d/+20,-8
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Previously, when `addon_tags` were changed, the response still used cached results based on the old value.

This commit fixes the issue by changing `addon_tags` from a query condition to a parameter.

It also enhances `addon_tags` to support category exclusion. `+xxx` means the result must include that category; `-xxx` means the result must not include it. A result must contain all `+xxx` tags and none of the `-xxx` tags.

Since javdb does not natively support exclusion filters, this is implemented by fetching videos in batches of 5 and filtering them manually. If the filtered results from a page are not enough to satisfy the limit, the next page will be requested until enough valid entries are collected.

Note: This may trigger javdb's anti-crawling mechanism more easily.

### DIFF
--- a/lib/routes/javdb/utils.ts
+++ b/lib/routes/javdb/utils.ts
@@ -32,132 +32,140 @@ const ProcessItems = async (ctx, currentUrl, title, excludeTags = new Set()) => 
         cookie && cookieJar.setCookie(cookie, rootUrl);
     }
 
-    let subject: string | undefined;
     const results = [] as DataItem[];
-    let page = 0;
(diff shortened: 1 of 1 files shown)


TonyRL:
- Highly overlapped with global [filtering](https://docs.rsshub.app/guide/parameters#filtering) parameter
- No page turn allowed

PR #19448, opened 2025-06-24
Exclude NSFW feeds when env variable DISABLE_NSFW is set to true

lawrenceching:

Close #

```routes
NOROUTE
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

In order to fulfill regulatory in some contries or regions, some feeds should be banned.
I add env variable `DISABLE_NSFW`, when it's "true",  NSFW feeds will be excluded from the build.


TonyRL (commented):
Reject the request in the registry when the env is set instead of modifying the build script.

lawrenceching:
> Reject the request in the registry when the env is set instead of modifying the build script.

Hi Tony, updated accordingly.
Let me know if it's what you want.

lawrenceching:
One question


```
namespaces = await import('../assets/build/routes.json');
```
The import wrap the content in default property. I printed logs for this. As you can see below, the "default" key is at the end of the line.
```
Object.keys(namespaces): a9vg,aamacau,********,default
```

So I have to make below change in order to fix my unit test failure:
```
if (namespaces.default) {
      namespaces = namespaces.default;
}
```

Do you think it's ok?

eslint.config.mjs
```diff
@@ -77,7 +77,7 @@ unicorn.configs.recommended,
         }],
 
         'default-case-last': 'error',
-        'no-console': 'error',
+        'no-console': 'off',
```
TonyRL: **DO NO** change lint rules. They're used to flag bad practises.

TonyRL (changes requested):

eslint.config.mjs
lawrenceching: It's edited for debugging purpose.
I will rollback this change at the final version

lawrenceching:
Hi @TonyRL , it seems everything is fine.  good to merge?

lawrenceching:
Hi @TonyRL @DIYgod , any update?

lib/registry.ts
```diff
@@ -48,6 +72,10 @@ switch (process.env.NODE_ENV) {
         }) as typeof modules;
 }
 
+if (process.env.DISABLE_NSFW === 'true') {
```
TonyRL: Use https://github.com/DIYgod/RSSHub/blob/6e202ac335b545abfb634d9e5af760a2e272af22/lib/config.ts#L75-L79 to handle the feature flag instead of reading `process.env.DISABLE_NSFW` directly

lawrenceching: done

lib/registry.ts
```diff
@@ -14,24 +14,44 @@ import logger from '@/utils/logger';
...
+            safe[key] = value;
+        }
+    }
+    return safe;
+}
+
 let modules: Record<string, { route: Route } | { namespace: Namespace }> = {};
-let namespaces: Record<
+
+export type RoutesType = Record<
```
TonyRL:
```suggestion
type RoutesType = Record<
```

PR #19852, opened 2025-08-17
fix(route): fix guid for Hacker News

wb14123:

Close #

```routes
/hackernews/index/comment
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Current guid for Hacker News has the number of comments in it when the type is not `sources`. This makes the guid change whenever a new comment is created under the post, which shouldn't happen, since the purpose of guid is to not duplicate the article when it is updated.

Quote from [RSS spec](https://www.rssboard.org/rss-2-0-1):

> A frequently asked question about `<guid>`s is how do they compare to
> `<link>`s. Aren't they the same thing? Yes, in some content systems, and
> no in others. In some systems, `<link>` is a permalink to a weblog item.
> However, in other systems, each `<item>` is a synopsis of a longer
> article, `<link>` points to the article, and `<guid>` is the permalink to
> the weblog entry. In all cases, it's recommended that you provide the
> guid, and if possible make it a permalink. This enables aggregators to
> not repeat items, even if there have been editing changes.

### DIFF
--- a/lib/routes/hackernews/index.ts
+++ b/lib/routes/hackernews/index.ts
@@ -79,7 +79,6 @@ async function handler(ctx) {
             item.upvotes = thing.next().find('.score').text().split(' point')[0];
 
             item.currentComment = thing.find('.comment').text();
-            item.guid = type === 'sources' ? item.guid : `${item.guid}${item.comments === 'discuss' ? '' : `-${item.comments}`}`;
 
             item.description = `<a href="${item.link}">Comments on Hacker News</a> | <a href="${item.origin}">Source</a>`;
 


wb14123:
Seems tests failed because of timeout?

TonyRL:
> ...when the type is not `sources`. This makes the guid change whenever a new comment is created under the post, which shouldn't happen...

This should be happened since when the type is either https://github.com/DIYgod/RSSHub/blob/715e9db0cf8d934615114fed1a73bb708b398c7e/lib/routes/hackernews/index.ts#L91 or https://github.com/DIYgod/RSSHub/blob/715e9db0cf8d934615114fed1a73bb708b398c7e/lib/routes/hackernews/index.ts#L125
the comments are RSS entries. When a new comment is posted, i.e., the "article" is updated, the `guid` should be updated accordingly.

FYI, [this is](https://github.com/DIYgod/RSSHub/blob/57b8489f5b6cc301933d32d8f6d9fc23808fb57f/docs/en/programming.md?plain=1#L326) corresponding docs which was lost during updates.

wb14123:
@TonyRL The point is, as quoted from the RSS spec above, even if the "article" is updated, the `guid` shouldn't be updated. Otherwise when there is a new comment for the HackerNews post, the RSS reader will make a duplicate entry for the same post since it thinks it's a totally different item.

PR #20053, opened 2025-09-15
feat(route)!: standardized the szse stock disclosure url

la3rence:

Close #

```routes
/szse/disclosure/listed/notice
/szse/disclosure/listed/notice/stock/000001
/szse/disclosure/listed/notice/stock/000002?beginDate=2025-01-01&endDate=2025-07-01
```

- [X] New Route / 新的路由
  - [X] Follows [Script Standard](https://docs.rsshub.app/joinus/advanced/script-standard) / 跟随 [路由规范](https://docs.rsshub.app/zh/joinus/advanced/script-standard)
- [X] [Date and time](https://docs.rsshub.app/joinus/advanced/pub-date) / [日期和时间](https://docs.rsshub.app/zh/joinus/advanced/pub-date)
  - [X] Parsed / 可以解析
  - [X] Correct time zone / 时区正确
- [ ] `Puppeteer`

## Note / 说明

This PR added support for specifying stock directly in the route path for `/szse/disclosure/listed/notice`, improving clarity and usability.
Also, it refactored parameter handling to use query strings for beginDate and endDate, removing the old query string parsing logic.

The old URL for fetching stock notice of 000001 is:

```
/szse/disclosure/listed/notice/stock=000001
```

The above URL format is not compliant with HTTP/URI RFCs, because `=` is not valid in the path segment (it should be used in query strings, which starts with `?`). This makes the route ambiguous and inconsistent with standard RESTful conventions. My FreshRSS instance rejects this feed due to this issue.

The new path format fixes this issue:

```
/szse/disclosure/listed/notice/stock/000001
```

I initially tried switching the URL to `?stock=000001`, but it seems like RSSHub’s caching mechanism does not support custom query parameters for route differentiation - `?stock=000002` will hit the cache `?stock=000001` (unless adding a `limit` query to bypass the cache). Because of this limitation, I had to introduce a new dedicated route with stock as a path parameter.

### DIFF
--- a/lib/routes/szse/disclosure/listed-notice.ts
+++ b/lib/routes/szse/disclosure/listed-notice.ts
@@ -24,20 +24,11 @@ function isValidDate(dateString: string): boolean {
 export const handler = async (ctx: Context): Promise<Data> => {
     const { category = '' } = ctx.req.param();
     const limit: number = Number.parseInt(ctx.req.query('limit') ?? '50', 10);
-    const query: string = ctx.req.param('query') ?? '';
     const queries: Record<string, string> = {
-        stock: '',
-        beginDate: '',
-        endDate: '',
+        stock: ctx.req.param('stock') ?? '',
+        beginDate: ctx.req.query('beginDate') ?? '',
+        endDate: ctx.req.query('endDate') ?? '',
     };
-    if (query) {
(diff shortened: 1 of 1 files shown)


la3rence:
Thanks for the comment. You're right about this. Encoding the `=` works.

PR #20100, opened 2025-09-19
feat(route): add Wikipedia current events

aavanian:
Add RSS feed for Wikipedia Portal:Current events providing the past 7 days of current events. Parses MediaWiki templates and converts wiki markup to HTML.

```routes
/wikipedia/current-events
```

- [x] New Route / 新的路由
  - [x] Parsed / 可以解析
- [ ] `Puppeteer`

### DIFF
--- a/lib/routes/wikipedia/current-events.ts
+++ b/lib/routes/wikipedia/current-events.ts
@@ -223,7 +223,7 @@ async function handler() {
         const cacheKey = `wikipedia:current-events:${pageName}`;
 
         try {
-            const wikitext = await cache.tryGet(cacheKey, async () => await fetchWikiRaw(pageName), 60 * 60 * 6); // Cache for 6 hours
+            const wikitext = await cache.tryGet(cacheKey, async () => await fetchWikiRaw(pageName), config.cache.contentExpire);
 
             // Parse the Current events template content
             const content = parseCurrentEventsTemplate(wikitext);
--- a/lib/routes/wikipedia/current-events.ts
+++ b/lib/routes/wikipedia/current-events.ts
@@ -1,0 +1,241 @@
+import { Route } from '@/types';
+import cache from '@/utils/cache';
+import got from '@/utils/got';
+import { parseDate } from '@/utils/parse-date';
+
+// Simple MediaWiki template parser for {{Current events}} template
+function parseCurrentEventsTemplate(wikitext: string): string | null {
+    // Look for {{Current events|content=...}} template
+    const templateMatch = wikitext.match(/\{\{Current events\s*\|[\s\S]*?content\s*=\s*([\s\S]*?)\}\}/);
+    if (!templateMatch) {
+        return null;
+    }
+
+    return templateMatch[1].trim();
+}
+
+// Wiki markup to HTML converter with proper list handling
+function wikiToHtml(wikitext: string): string {
+    let html = wikitext;
+
+    // Convert wiki links [[Link|Text]] or [[Link]]
+    html = html.replaceAll(/\[\[([^|\]]+)\|([^\]]+)\]\]/g, '<a href="https://en.wikipedia.org/wiki/$1">$2</a>');
+    html = html.replaceAll(/\[\[([^\]]+)\]\]/g, '<a href="https://en.wikipedia.org/wiki/$1">$1</a>');
+
+    // Convert external links [URL Text] or [URL]
+    html = html.replaceAll(/\[([^\s\]]+)\s+([^\]]+)\]/g, '<a href="$1">$2</a>');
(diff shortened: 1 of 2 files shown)


dzx-dzx:
What about https://en.wikipedia.org/w/index.php?title=Portal:Current_events&action=history ?

aavanian:
If you mean the 2 feeds on the page you linked, those are editing feeds, showing individual edit actions, not the actual page.
Same for the native feed on https://en.wikipedia.org/wiki/Portal:Current_events, it's not the content but the edits. 
That's why I wrote this one.

lib/routes/wikipedia/current-events.ts
```diff
@@ -0,0 +1,260 @@
...
+}
+
+async function fetchWikiRaw(pageName: string): Promise<string> {
+    const url = `https://en.wikipedia.org/wiki/${pageName}`;
+    const response = await got(url, {
+        searchParams: {
+            action: 'raw',
+        },
+        headers: {
+            'User-Agent': 'RSSHub/1.0 (+https://github.com/DIYgod/RSSHub)',
```
TonyRL: Please use the provided value https://github.com/DIYgod/RSSHub/blob/380874aca8abfff9b1aad2cb77b50ecccbc77be5/lib/config.ts#L440 instead

lib/routes/wikipedia/current-events.ts
```diff
@@ -0,0 +1,260 @@
...
+        return date;
+    });
+
+    // Fetch all pages in parallel to avoid await-in-loop
+    const fetchPromises = dates.map(async (date) => {
+        const pageName = getCurrentEventsDatePath(date);
+        const cacheKey = `wikipedia:current-events:${pageName}`;
+
+        try {
+            const wikitext = await cache.tryGet(cacheKey, async () => await fetchWikiRaw(pageName), 60 * 60 * 6); // Cache for 6 hours
```
TonyRL: Do not hard code the cache duration. It should be managed by https://github.com/DIYgod/RSSHub/blob/380874aca8abfff9b1aad2cb77b50ecccbc77be5/lib/config.ts#L501

TonyRL: Please use the [MediaWiki API](https://www.mediawiki.org/wiki/API:Action_API) instead. This follows the etiquette when using the API and allows RSSHub to [send only one request](https://www.mediawiki.org/wiki/API:Etiquette#Request_limit) to Wikipedia.

aavanian: fixed with 6f56be3

lib/routes/wikipedia/current-events.ts
aavanian: fixed with 10ed58e

lib/routes/wikipedia/current-events.ts
aavanian: Thanks! Unfortunately, their parse API is single document only. Multi-document request only work for revision in wikitext format so I had to keep the wikitext -> html conversion code.

In my tests, a single request is sufficient to recover the 7 days but pagination is implemented in case that changes.

fixed with 1f332bc

lib/routes/wikipedia/current-events.ts
```diff
@@ -0,0 +1,379 @@
...
+            ...continueParams,
+        };
+
+        // eslint-disable-next-line no-await-in-loop
+        const response = await got(url, {
+            searchParams,
+            headers: {
+                'User-Agent': config.trueUA,
+            },
+        });
```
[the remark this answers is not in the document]
aavanian: It's the proper way in that case though. We do want the calls to be sequential and in any case they have to be since it's for pagination (the next request depends on the result of the preceding).

TonyRL (commented):
Replace all `\n` with `<br>` instead.

aavanian:
> Replace all `\n` with `<br>` instead.

this one:

https://github.com/DIYgod/RSSHub/blob/1f332bcabf5b99a7f325d23fc73d44015d256218/lib/routes/wikipedia/current-events.ts#L82

is necessary for further processing (and removed in the process).

this one:

https://github.com/DIYgod/RSSHub/blob/1f332bcabf5b99a7f325d23fc73d44015d256218/lib/routes/wikipedia/current-events.ts#L140

is purely cosmetic and could just have been joined with `''` (done in f6cecdbb). Using `<br>` would break the formatting.

Considering how the "Current Events" portal is formatted (emphasized sections with content of each being strictly bullet points), there's no semantically relevant line breaks (`\n\n` in wikitext) that are not taken care of by the current implementation. 

Note the parsing and conversion of the wikitext to html implemented here is very limited and ad-hoc. There are two alternatives, adding a parser dependency (👎 ) or requesting an rendered page (request is slower and we still need to process the html). I reviewed the different choices and added that review in a code comment.

PR #20151, opened 2025-09-25
feat(route): add /generic_proxy to proxy arbitrary http/https resources

zaytsevsgit:
Adds a new route: /generic_proxy/:url{.+}
Purpose: acts as a file proxy. Given a URL-encoded absolute http/https URL, returns the upstream response (status, key headers, body).

Implementation details:
GET-only enforcement
URL decoding and scheme validation (http/https only)
Upstream fetch via @/utils/got with default config (config.requestTimeout, config.ua) overridable by env (GENERIC_PROXY_TIMEOUT, GENERIC_PROXY_MAX_SIZE, GENERIC_PROXY_UA)
Propagates key headers (Content-Type, Content-Length, Cache-Control, ETag, Last-Modified, Content-Disposition)
Response size limit (default 10 MB) to prevent excessive memory usage
Tests included: status forwarding (404), small binary fetch (10 bytes), 405 for non-GET, 400 for unsupported scheme

Checklist:
[x] Route file under lib/routes/
[x] Declarative export const route used
[x] No changes to lib/router.js (respect “DO NOT ADD ANY NEW ROUTES HERE”)
[x] Tests pass locally (vitest)
[x] Lints pass


zaytsevsgit:
"fix(route): use kebab-case filenames and rename catch parameter"

zaytsevsgit:
check

PR #20290, opened 2025-10-13
fix(route/artstation): csrf token

### LINKED ISSUE #20289: artstation fails to get csrf token
### Routes

```routes
/artstation/:handle
```

### Full routes

```fullroutes
/artstation/wlop
```

### Related documentation

https://docs.rsshub.app/routes/picture#artstation

### What is expected?

Any given artstation handle should successfully return the posts for the user.

The request fails with the error `TypeError: Cannot read properties of undefined (reading 'set-cookie')`

### Deployment information

Self-hosted

OS: Linux, Node: v22.17.0, Docker: 28.0.1

### Additional info

Image

The error occurs because got.post does not return (used to?) headers, so reading the csrf token fails.

https://github.com/DIYgod/RSSHub/blob/07d16ae7ba8d0d8686f2d3c011748f3995ff904d/lib/routes/artstation/user.ts#L47

EldritchGarden:

Close #20289 

```routes
/artstation/wlop
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

- Fixed fetch for CSRF token

### DIFF
--- a/lib/routes/artstation/user.ts
+++ b/lib/routes/artstation/user.ts
@@ -2,6 +2,7 @@ import { Route } from '@/types';
 
 import cache from '@/utils/cache';
 import got from '@/utils/got';
+import ofetch from '@/utils/ofetch';
 import { parseDate } from '@/utils/parse-date';
 import path from 'node:path';
 import { art } from '@/utils/render';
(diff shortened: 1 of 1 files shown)


lib/routes/artstation/user.ts
```diff
@@ -44,7 +44,7 @@ async function handler(ctx) {
         const tokenResponse = await got.post('https://www.artstation.com/api/v2/csrf_protection/token.json', {
             headers,
         });
-        return tokenResponse.headers['set-cookie'][0].split(';')[0].split('=')[1];
+        return tokenResponse.data.public_csrf_token;
```
TonyRL: Please double-check the value. Sequential requests use `PRIVATE-CSRF-TOKEN` from the set-cookie header which has a difference value than `public_csrf_token`.

EldritchGarden: You're right, I didn't catch that. Switched back to using the set-cookie header to get the private csrf token value

PR #20401, opened 2025-10-31
fix(youtube): fix embed parameter handling

jean-jacket:

Close #

```routes
/youtube/user/@mkbhd/videos
/youtube/channel/UC7yZ6keOGsvERMp2HaEbbXQ/embed?=false
```

- [ ] New Route / 新的路由
  - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明
Using `params.get` means embed is always `true` even if you add the parameter.

### DIFF
--- a/lib/routes/youtube/channel.ts
+++ b/lib/routes/youtube/channel.ts
@@ -55,7 +55,7 @@ async function handler(ctx) {
     const params = new URLSearchParams(routeParams);
 
     // Get embed parameter
-    const embed = !params.get('embed');
+    const embed = !params.has('embed');
 
     // Get filterShorts parameter (default to true if not specified)
     const filterShortsStr = params.get('filterShorts');
--- a/lib/routes/youtube/user.ts
+++ b/lib/routes/youtube/user.ts
@@ -51,7 +51,7 @@ async function handler(ctx) {
     const params = new URLSearchParams(routeParams);
 
     // Get embed parameter
-    const embed = !params.get('embed');
+    const embed = !params.has('embed');
 
     // Get filterShorts parameter (default to true if not specified)
     const filterShortsStr = params.get('filterShorts');


TonyRL:
Unnecessary changes. `embed` does get flipped when it's **properly** passed.
image

You can read more on this in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) and [RFC 3987](https://datatracker.ietf.org/doc/html/rfc3987)

jean-jacket:
I see the confusion now. 

The "proper" way to pass it is like this:
`http://localhost:1200/youtube/channel/UC7yZ6keOGsvERMp2HaEbbXQ/embed=false`

However this is unusual and the documentation isn't clear. 

Query parameters are usually passed like this:
`http://localhost:1200/youtube/channel/UC7yZ6keOGsvERMp2HaEbbXQ?embed=false`

(See the RFC)

TonyRL:
They are not query parameters at all. Also, passing parameter in path does follow the definition of `path` in [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986).

As you can see from [rfc3986 1.1.2](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.2), it shows an example containing `=` in path though the protocol is LDAP instead of http.
> ldap://[2001:db8::7]/c=GB?objectClass?one

[rfc3986 2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) defines `=` belongs to `sub-delims`
> sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
/ "*" / "+" / "," / ";" / "="

[rfc3986 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) which defines the syntax of path.
> path          = path-abempty
> path-abempty  = *( "/" segment )
> segment       = *pchar
> pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"

PR #20668, opened 2025-12-09
fix(cache): always use JSON.stringify when storing values to cache

### LINKED ISSUE #20667: Cache incorrectly parsed as JSON value
### Routes

```routes
/mastodon/acct/:acct/statuses/:only_media?
```

### Full routes

```fullroutes
/mastodon/acct/linuxmagazine@fosstodon.org/statuses
```

### Related documentation

https://docs.rsshub.app/joinus/advanced/use-cache

### What is expected?

Numeric strings are fetched as strings

Numeric strings are not quoted when stored, and are later parsed as numbers, and precision is lost

https://github.com/DIYgod/RSSHub/blob/781da24210043ad9a8de480567ed2862403770fd/lib/utils/cache/index.ts#L80-L97

### Deployment information

Self-hosted

_No response_

### Additional info

If some values need to be serialized as JSON, then all types should.

#20663 was an incorrect fix. I noticed that the error was caused by cache, but didn't identify the root cause.

ouuan:

Close #20667

```routes
/mastodon/acct/linuxmagazine@fosstodon.org/statuses
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

I don't know why it was implemented as it was, and I'm not sure if this fix has any unexpected side effects.

### DIFF
--- a/lib/utils/cache/index.ts
+++ b/lib/utils/cache/index.ts
@@ -36,12 +36,7 @@ if (config.cache.type === 'redis') {
         }
     };
     globalCache.set = (key, value, maxAge = config.cache.routeExpire) => {
-        if (!value || value === 'undefined') {
-            value = '';
-        }
-        if (typeof value === 'object') {
-            value = JSON.stringify(value);
-        }
+        value = JSON.stringify(value);
         if (key && memoryCache) {
             return memoryCache.set(key, value, { ttl: maxAge * 1000 });
         }
--- a/lib/utils/cache/redis.ts
+++ b/lib/utils/cache/redis.ts
@@ -58,12 +58,7 @@ export default {
         if (!status.available || !clients.redisClient) {
             return;
         }
-        if (!value || value === 'undefined') {
-            value = '';
-        }
-        if (typeof value === 'object') {
-            value = JSON.stringify(value);
-        }
+        value = JSON.stringify(value);
         if (key) {
             if (maxAge !== config.cache.contentExpire) {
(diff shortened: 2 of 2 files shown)


lib/utils/cache/index.ts
```diff
@@ -36,12 +36,7 @@ if (config.cache.type === 'redis') {
         }
     };
     globalCache.set = (key, value, maxAge = config.cache.routeExpire) => {
-        if (!value || value === 'undefined') {
-            value = '';
-        }
-        if (typeof value === 'object') {
-            value = JSON.stringify(value);
-        }
```
TonyRL, line 44: They have been there 7 years ago (a1f060f3d2205fbe5cfc1672ad2bb0212e350c65, 2257875db752c133c4b38424599e115cf959c4ff), and changing them obviously breaks the tests.

ouuan:
I see. Some routes have already called `JSON.stringify` in `cache.set`. Maybe the correct way of fixing this is to call `JSON.stringify` in routes.

ouuan:
I considered some other fixes, but any routes might be relying on the current behavior, so we can only fix them case by case 🫠

ouuan:
Oh, just noticed that the value is returned as-is without `JSON.parse` in `tryGet`:

https://github.com/DIYgod/RSSHub/blob/781da24210043ad9a8de480567ed2862403770fd/lib/utils/cache/index.ts#L92-L97

So maybe it's safe to assume the `value` here is not JSON stringified.

PR #20671, opened 2025-12-10
fix(cache): call JSON.stringify when storing value in cache.tryGet

### LINKED ISSUE #20667: Cache incorrectly parsed as JSON value

ouuan:

Close #20667

```routes
/mastodon/acct/linuxmagazine@fosstodon.org/statuses
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

This replaces #20668.

https://github.com/DIYgod/RSSHub/blob/e5ae378fe955c82780d84a6d4699bdf13e5834df/lib/utils/cache/index.ts#L79-L97

The value from `getValueFunc` is returned as-is without `JSON.parse`, so we may assume that it's not JSON-stringified inside `getValueFunc`. Otherwise, the caller already needs to distinguish between values that are either JSON-parsed or not.


ouuan:
P.S. It's troublesome to fix this in each feed by adding `JSON.stringify` in `getValueFunc`, as the value from `getValueFunc` is returned as-is.

ouuan:
Can we merge this or is there any concern?

ouuan:
Please review this

PR #20597, opened 2025-11-29
feat(nix): add Nix flake and NixOS module support

DzmingLi:
## Involved Issue / 该 PR 相关 Issue

N/A

## Example for the Proposed Route(s) / 路由地址示例

```routes
NOROUTE
```

## New RSS Route Checklist / 新 RSS 路由检查表

Not applicable - this PR adds Nix infrastructure support.

## Note / 说明

This PR adds comprehensive Nix support for RSSHub, making deployment on NixOS significantly easier.

### What's included:

1. **Nix Flake (`flake.nix`)**:
   - NixOS service module for declarative deployment
   - Package definition for building RSSHub with Nix
   - Overlay for easy integration

2. **Development Environment (`devenv.nix`)**:
   - Reproducible development environment with Node.js 22 and pnpm
   - Optional Redis service integration
   - Convenient scripts: `rsshub-dev`, `rsshub-build`, `rsshub-start`, `rsshub-test`
   - Pre-commit hooks integration

3. **Direnv Integration (`.envrc`)**:
   - Automatic environment loading when entering the project directory

### Benefits for NixOS users:

- **Declarative deployment**: Configure RSSHub in NixOS configuration.nix
- **Reproducible builds**: Lock dependencies with flake.lock
- **Service management**: Systemd service with security hardening
- **Redis integration**: Built-in Redis caching support
- **Zero-config development**: Just `direnv allow` to get started

This brings RSSHub deployment on NixOS to the same ease-of-use level as Docker, while maintaining the declarative and reproducible nature of NixOS.

PR #20681, opened 2025-12-12
fix(.envrc): add --no-pure-eval for devenv flake compatibility

mefengl:
Follow-up to #20597.

When entering the directory with direnv, devenv fails with:
```
error: Failed assertions:
- devenv was not able to determine the current directory.
```

devenv requires `--no-pure-eval` to determine project root via `builtins.getEnv "PWD"` in flake mode.

Ref: 
- https://devenv.sh/guides/using-with-flakes/#automated-shell-switching
- cachix/devenv#1461

cc @DzmingLi

also add `.direnv`, `.devenv` and `.pre-commit-config.yaml` to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@
 .yarn
 .yarnrc.yml
 .pnp*
+.direnv
+.devenv
+.pre-commit-config.yaml
 
 *.swp
 *.iml


.gitignore
```diff
@@ -35,3 +35,6 @@ package-lock.json
 # pnpm-lock.yaml
 yarn.lock
 yarn-error.log
+.devenv
```
TonyRL: Dot files are added in the first group obviously.

mefengl: done

.gitignore
```diff
@@ -12,6 +12,9 @@
 .yarn
 .yarnrc.yml
 .pnp*
+.direnv
```
TonyRL, line 15: And follows alphabetical order obviously.

PR #21164, opened 2026-02-16
feat(route/twitter): support parsing tweet threads from web-api response

mindt102:

Close #

```routes
/twitter/user/_RSSHub
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Some tweets from user timeline are not being parsed as they are grouped in a thread like in the example below

image

I added the entryId "profile-conversation-" of these tweet to the filter and modified the code to handle these missing tweets

image

### DIFF
--- a/lib/routes/twitter/api/web-api/utils.ts
+++ b/lib/routes/twitter/api/web-api/utils.ts
@@ -311,7 +311,7 @@ export function gatherLegacyFromData(entries: any[], filterNested?: string[], us
             const itemTweets = content?.items
                 ?.map((i) => i?.item?.itemContent?.tweet_results?.result)
                 .filter(Boolean);
-	
+
             if (itemTweets?.length > 0) {
                 logger.debug('Found a tweet thread');
                 for (let t of itemTweets) {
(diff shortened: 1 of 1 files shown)


TonyRL (commented):
~These are replies and should be only added when~ https://github.com/DIYgod/RSSHub/blob/4b761522f2f00c098d1324a5b01a7ee98aa1b91d/lib/routes/twitter/namespace.ts#L27
~`includeReplies` is true~

You're right. https://github.com/DIYgod/RSSHub/blob/ba0f1ad2920df4b71f2244471d21aaa73779b01e/lib/routes/twitter/api/mobile-api/api.ts#L274-L279

mindt102:
The tweets in my screenshot were from the Posts section in a user timeline. I understand that tweet (2) could be considered a reply, but tweet (1) was not a reply and it was not included in the RSS feed because it had a different id ("profile-conversation-").

image

My current understanding from reading the code is that:
- if `includeReplies=false`, the tweets in the Posts sections would be query using the `UserTweets` API
- if `includeReplies=true`, the tweets in the Replies sections would be query using the `UserTweetsAndReplies` API. 

The tweets from my screenshot are already included from the result when calling `UserTweets` API, so I think it might be useful to parse it and add to the RSS feed so that no posts by the target user is missing.

As a side note, I tried using the `includeReplies=true` option and it keeps getting a 404 status code. It seems to be a changes from Twitter API that cause this issue and make that option unusable at the moment.

PR #21253, opened 2026-02-28
fix(zhihu): use v4 API for daily route

choury:
## Involved Issue

Close #

## Example for the Proposed Route(s)

```routes
/zhihu/daily
```

## New RSS Route Checklist

- [ ] New Route
- [x] Bug fix

### Problem

The `/zhihu/daily` route uses v7 API (`/api/7/story/:id`) to fetch story details, which no longer returns `publish_time`, causing `pubDate` to be `Invalid Date` for all items.

Additionally, some stories on the homepage may return HTTP 500 from the story detail API, which causes `Promise.all` to reject and the entire route to return 503.

### Fix

- Keep HTML scraping of `daily.zhihu.com` homepage for story list (30 items via `.box` elements)
- Switch story detail API from v7 to v4 (`/api/4/story/:id`), which returns `publish_time` as a Unix timestamp
- Add try-catch around individual story fetches so a single failed story does not crash the entire route
- Add `image` field from story detail response

### Verified locally

Compiled, restarted, and tested: 29 items returned with valid pubDate (1 story skipped due to server-side HTTP 500 on story/9787342).

### DIFF
--- a/lib/routes/zhihu/daily.ts
+++ b/lib/routes/zhihu/daily.ts
@@ -2,6 +2,7 @@ import { load } from 'cheerio';
 
 import type { Route } from '@/types';
 import cache from '@/utils/cache';
+import logger from '@/utils/logger';
 import ofetch from '@/utils/ofetch';
 import { parseDate } from '@/utils/parse-date';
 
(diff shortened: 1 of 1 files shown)


TonyRL (changes requested):
This greatly reduces the number of feed entries from 30 to 4.

> the `.box` elements with story links no longer exist

Please double-check it in the DevTool from the browser manually. 

<details><summary>Screenshot</summary>

Image
</details>

PR #21542, opened 2026-03-28
feat(route/locals): add content feed

luckycold:

No issue involved. This was just something I wanted and thought others might like too.

```routes
/locals/content/mattfradd
/locals/content/mattfradd/video
/locals/content/mattfradd/plus
/locals/content/mattfradd/nonplus/audio
/locals/content/mattfradd/podcast
```

- [x] New Route / 新的路由
  - [x] Follows Script Standard / 跟随 路由规范
- [x] Date and time / 日期和时间
  - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

This route uses `LOCALS_SESSION` route-specific configuration and Locals' authenticated server functions instead of scraping the rendered page.

### Implementation details

- **Auth**: Uses the `locals2.v3.session` cookie through `LOCALS_SESSION`
- **Data source**: Calls Locals' internal content server functions on `locals.com` for stable content feed data
- **Filtering**: Supports merged content, `plus` / `nonplus`, and media filters for `video`, `live`, `audio` / `podcast`, `article`, and `document`
- **Validation**: Verified locally against `mattfradd` for `/locals/content/mattfradd`, `/locals/content/mattfradd/video`, `/locals/content/mattfradd/plus`, `/locals/content/mattfradd/nonplus/audio`, and `/locals/content/mattfradd/podcast`


luckycold:
Verified locally again after the latest update: Locals content items now expose thumbnails in three ways for better reader compatibility:

- `image`
- `itunes:image`
- an `` at the start of `description`

I re-tested this on a fresh local RSSHub instance with `LOCALS_SESSION`, including:

- `/locals/content/mattfradd`
- `/locals/content/mattfradd/video`
- `/locals/content/mattfradd/podcast`

The latest change also fixes image extraction for Locals video previews and podcast/audio artwork, so screenshots should now show the thumbnails correctly in the rendered feed output.

luckycold:
The failing route automation is expected here because this route requires route-specific configuration (`LOCALS_SESSION`).

I verified the route locally with a real authenticated `locals2.v3.session` cookie and a local RSSHub instance. These all return successfully:
- /locals/content/mattfradd
- /locals/content/mattfradd/video
- /locals/content/mattfradd/plus
- /locals/content/mattfradd/nonplus/audio
- /locals/content/mattfradd/podcast

This route uses Locals' authenticated server functions on `locals.com` rather than scraping rendered page HTML, so the content feed data is coming from the same internal source the site uses.
I attached some screenshots showing that it is actually parsing the RSS correctly though and I also included how it is shown in my local RSS reader.

image
image
image

luckycold:
I noticed that it wasn't using proper thumbnails from the actual post thumbnail. I updated it to use the poster thumbnails instead.

luckycold:
Alright. I further fixed it to match codex's comments. And here's a view of what the fixed thumbnails look like now (reflects what's shown on locals):
image
image

lib/routes/locals/feed.ts
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    return {
+        contentType: contentType ? contentTypeMap[contentType] : undefined,
+        filter,
+    };
+}
+
+async function requestServerFunction<T>(session: string, id: string, key: string, args: unknown[]) {
+    const response = await fetch('https://locals.com/_server', {
```
TonyRL: Use ofetch instead node's built-in fetch

lib/routes/locals/feed.ts
```diff
@@ -0,0 +1,440 @@
...
+                existing.itunes_item_image = existing.itunes_item_image || item.itunes_item_image;
+            }
+        }
+    }
+
+    return [...items.values()].toSorted((a, b) => {
+        const aTime = a.pubDate ? new Date(a.pubDate).getTime() : 0;
+        const bTime = b.pubDate ? new Date(b.pubDate).getTime() : 0;
+        return bTime - aTime;
+    });
```
TonyRL: Unnecessary sorting. RSSHub will do this automatically. https://github.com/DIYgod/RSSHub/blob/63633560d1a5f8c37cbbed551da80d497b3ab442/lib/middleware/parameter.ts#L85

lib/routes/locals/feed.ts
```diff
@@ -0,0 +1,440 @@
...
+};
+
+function getCookieHeader(session: string) {
+    return [`locals2.v3.session=${session}`, 'locals.preferLocals2=false', 'locals2.v3.locale.actual=en-us%2Cen', 'locals2.v3.locale.inferred=en'].join('; ');
+}
+
+function getRequestHeaders(session: string) {
+    return {
+        Cookie: getCookieHeader(session),
+        'User-Agent': config.trueUA,
```
TonyRL: This route works fine with auto-injected UA which is Chrome on mac. Only use this UA when the default one doesn't work.

lib/routes/locals/feed.ts
```diff
@@ -0,0 +1,440 @@
...
+            thumb?: string;
+        };
+    };
+    hashtag: string;
+    id: number;
+    title: string;
+};
+
+export const route: Route = {
+    path: ['/content/:community', '/content/:community/:option1', '/content/:community/:option1/:option2'],
```
TonyRL: Provide only 1 route path instead of multiple

lib/config.ts
TonyRL, line 1: `L` comes before `M`

luckycold: Not sure what happened there. I thought I fixed this. My bad.

lib/routes/locals/feed.ts
luckycold: Done

lib/routes/locals/feed.ts
luckycold: Done and noted for the future.

lib/routes/locals/feed.ts
luckycold: Done and noted.

lib/config.ts
luckycold: Done as well btw.

luckycold:
Made the quick fixes. I also sorted SECURITY_KEY which seems it was out of order as well. If you'd like me to undo it I can revert. Was bugging me after I began to see it.

lib/routes/locals/feed.ts
luckycold: Done as well.

lib/routes/locals/feed.ts
```diff
@@ -0,0 +1,436 @@
...
+};
+
+export const route: Route = {
+    path: '/content/:community/:option1?/:option2?',
+    categories: ['social-media'],
+    example: '/locals/content/mattfradd/video',
+    parameters: {
+        community: 'Community slug from `locals.com/:community/feed?mode=content`',
+        option1: 'Optional filter or content type. Filters: `plus`, `nonplus`. Content types: `video`, `live`, `audio`, `podcast`, `article`, `document`, `pdf`',
+        option2: 'Optional content type when `option1` is a filter',
```
TonyRL:
```suggestion
        option1: 'Filter or content type. Filters: `plus`, `nonplus`. Content types: `video`, `live`, `audio`, `podcast`, `article`, `document`, `pdf`',
        option2: 'Content type when `option1` is a filter',
```
The necessity of a parameter will be generated automatically based on `path`

PR #21957, opened 2026-05-09
fix(twitter): handle TweetPreviewDisplay for subscriber-only posts

Jsongcloud:

Close #

```routes
NOROUTE
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Fix subscriber-only posts (`TweetPreviewDisplay`) not appearing in `/twitter/user/:id` feed.

The X API returns subscriber-only posts with `__typename: "TweetPreviewDisplay"`, which has a different structure from regular `Tweet` objects. The parser silently dropped these entries because the type was not handled.

Added a handler in both `web-api/utils.ts` and `mobile-api/api.ts`. The check must occur before the `tweet = tweet.tweet` reassignment. Preview text is included in the feed with a 🔒 prefix.

### DIFF
--- a/lib/routes/twitter/api/mobile-api/api.ts
+++ b/lib/routes/twitter/api/mobile-api/api.ts
@@ -145,7 +145,34 @@ function gatherLegacyFromData(entries, filterNested, userId) {
     for (const entry of filteredEntries) {
         if (entry.entryId) {
             const content = entry.content || entry.item;
-            let tweet = content?.content?.tweetResult?.result || content?.itemContent?.tweet_results?.result;
+let tweet = content?.content?.tweetResult?.result || content?.itemContent?.tweet_results?.result;
+            // 处理订阅者专属预览帖子（必须在 tweet.tweet 替换之前判断）
+            if (tweet?.__typename === 'TweetPreviewDisplay') {
+                const preview = tweet.tweet;
+                if (preview?.rest_id) {
+                    const userResult = preview.core?.user_results?.result;
+                    const screenName = userResult?.core?.screen_name ?? '';
+                    const fakeLegacy = {
+                        id_str: preview.rest_id,
+                        full_text: `🔒 [Subscribers Only] ${preview.text ?? ''}`,
+                        created_at: preview.created_at ?? '',
+                        entities: { urls: [], hashtags: [], symbols: [], user_mentions: [] },
+                        user_id_str: userResult?.rest_id ?? '',
+                        user: {
+                            name: userResult?.core?.name ?? screenName,
+                            screen_name: screenName,
+                            profile_image_url_https: userResult?.avatar?.image_url ?? '',
+                        },
+                        favorite_count: preview.favorite_count ?? 0,
+                        reply_count: preview.reply_count ?? 0,
+                        retweet_count: preview.retweet_count ?? 0,
+                    };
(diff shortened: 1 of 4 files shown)


lib/routes/twitter/api/mobile-api/api.ts
```diff
@@ -146,6 +146,32 @@ function gatherLegacyFromData(entries, filterNested, userId) {
         if (entry.entryId) {
             const content = entry.content || entry.item;
             let tweet = content?.content?.tweetResult?.result || content?.itemContent?.tweet_results?.result;
+            // Handle subscriber-only preview posts (must check before tweet.tweet reassignment)
+            if (tweet?.__typename === 'TweetPreviewDisplay') {
```
TonyRL: Have you tested the mobile API? Could you provide a screenshot of RSS output using mobile API?

Jsongcloud: After reviewing the codebase, I found that mobile-api is currently commented out in lib/routes/twitter/api/index.ts and not in use. I have reverted the lib/routes/twitter/api/mobile-api/api.ts changes. The fix is now only applied to lib/routes/twitter/api/web-api/utils.ts which is the active implementation.

The fix is confirmed working - subscriber-only posts (TweetPreviewDisplay) appear correctly with 🔒 prefix in the RSS feed as shown in the screenshot.
Twitter_sub

lib/routes/twitter/api/web-api/utils.ts
```diff
@@ -338,6 +338,28 @@ export function gatherLegacyFromData(entries: any[], filterNested?: string[], us
...
             const content = entry.content || entry.item;
             let tweet = content?.content?.tweetResult?.result || content?.itemContent?.tweet_results?.result;
+            // Handle subscriber-only preview posts (must check before tweet.tweet reassignment)
+            if (tweet?.__typename === 'TweetPreviewDisplay') {
+                const preview = tweet.tweet;
+                if (preview?.rest_id) {
+                    const userResult = preview.core?.user_results?.result;
+                    const fakeLegacy: any = {
+                        id_str: preview.rest_id,
+                        full_text: `🔒 [Subscribers Only] ${preview.text ?? ''}`,
```
TonyRL: Follow the convention of https://github.com/DIYgod/RSSHub/blob/238226edda3dfcc948d6531971efaaf890865d0d/lib/routes/twitter/namespace.ts#L17-L18 to provide options that allow users to choose using "🔒" instead of "Subscribers only".

Jsongcloud: Currently I'm using `🔒 [Subscribers Only]` as the prefix. Should I split this into two options similar to `showEmojiForRetweetAndReply` and `showSymbolForRetweetAndReply`? For example:
- `showEmojiForSubscriberOnly` (default `true`): use "🔒" prefix
- `showSymbolForSubscriberOnly` (default `true`): use "[Subscribers Only]" prefix

Or should only one of them be kept and made configurable?

TonyRL: > Should I split this into two options similar to showEmojiForRetweetAndReply and showSymbolForRetweetAndReply?

Yes. However, they should not be true at the same time by default.

PR #21962, opened 2026-05-11
feat(cna): cleanup extracted full texts

yan12125:

N/A

```routes
/cna/aipl
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Here is a page with description-box:
https://www.cna.com.tw/news/aipl/202605070117.aspx

Explanation for the term 憲法第130條 is injected in the middle of a paragraph and makes that paragraph hard to read. This change fixes that.

### DIFF
--- a/lib/routes/cna/utils.ts
+++ b/lib/routes/cna/utils.ts
@@ -14,6 +14,13 @@ export async function getFullText(item) {
     const content = load(detailResponse.data);
     content('div.SubscriptionInner').remove();
     content('.gmailNews').remove();
+
+    // Those boxes are for explaining terms. They are injected inline and interrupt reading.
+    // If readers want to learn about terms, they can learn more online.
+    content('.dictionary-box').remove();
+    // Those are for "延伸閱讀" links, which are not part of the main article.
+    content('.moreArticle').remove();
+
     const topImage = content('.fullPic').html();
 
     item.description = (topImage === null ? '' : topImage) + content('.paragraph').eq(0).html();


lib/routes/cna/utils.ts
```diff
@@ -14,6 +14,15 @@ export async function getFullText(item) {
...
     content('div.SubscriptionInner').remove();
     content('.gmailNews').remove();
+
+    // Those boxes are for explaining terms. They are injected inline and interrupt reading.
+    // If readers want to learn about terms, they can learn more online.
+    content('.dictionary-box').remove();
+    // Those are for separating "延伸閱讀" links. On web pages, those links have style "display:block;".
+    // However, some RSS readers, such as TT-RSS, sanitize HTML aggressively and remove <style> tags and style= attributes.
+    // As a result, a wrapping div is needed.
+    content('.moreArticle-link').wrap('<div></div>');
```
TonyRL: Remove `.moreArticle` as well

yan12125: Do you mean removing the entire moreArticles block and all links inside it? I prefer to keep them, as those links are helpful for understanding the whole picture from relevant events.

TonyRL: Yes, those content aren't the main article.

yan12125: I agree that those links are not part of the main article, while I believe keeping them is better for cna. Otherwise, some useful information is stripped away compared to web pages. https://rsshub.netlify.app/joinus/new-rss/start-code#better-reading-experience mentions an example that "will have a similar reading experience to the original website" using full article extraction, and my idea roughly follows that spirit.

Ideally, the description box is also preserved, while I cannot figure out a way to represent it properly, particularly in restrictive RSS readers like TT-RSS, so I remove it to make the reading experience closer to web pages - an uninterrupted reading flow.

yan12125: Hi, could we move this forward? I'd like to confirm if I misunderstand the goal of full article extraction or not. I will update this PR as you suggested if I did misunderstand something.

TonyRL: You've already answered it yourself.

> goal of full **article** extraction

> those links are **not** part of the main article

yan12125: Alright, updated

Issue #22136, opened 2026-05-30
twitter: /twitter/user/* return 503 (TypeError in gql-id-resolver) + /twitter/keyword/* return 404 (regression on master 2026-05-29)


Dinglingyun2022:
### Routes

```routes
/twitter/user/:id
/twitter/keyword/:keyword
/twitter/user/:id/with_replies
```

### Full routes

```fullroutes
/twitter/user/dscassoc
/twitter/keyword/oled
/twitter/user/dscassoc/with_replies
```

### Related documentation

https://docs.rsshub.app/routes/social-media#twitter-x

### What is expected?

Returns an RSS XML containing the latest tweets for the given user / keyword search (HTTP 200 + valid `<rss>...</rss>` body).

### What is actually happening?

All `/twitter/*` routes return **HTTP 503**, in two distinct error patterns:

**Pattern 1 — `/twitter/user/:id` and `/twitter/user/:id/with_replies` crash in `gql-id-resolver`:**

```
info: <-- GET /twitter/user/dscassoc
info: twitter gql-id-resolver: fetching fresh query IDs from Twitter JS bundles
error: undefined
error: Error in /twitter/user/dscassoc: TypeError: Cannot read properties of undefined (reading 'replace')
info: --> GET /twitter/user/dscassoc 503 14s
```

The crash occurs inside the `gql-id-resolver` module — the dynamic GraphQL query ID resolver introduced by PR #21544. After PR #21787 (2026-05-21) added `X-Client-Transaction-Id` for `UserTweetsAndReplies`, X likely modified the JS bundle structure / URL pattern again, making the regex extract `undefined`, and the subsequent `.replace()` call throws.

The duplicate-detection bot on the previous filing (#22135) explicitly confirmed: *"For the /twitter/user/\* gql-id-resolver TypeError crash, no obvious duplicates found — this appears to be a new regression."*

**Pattern 2 — `/twitter/keyword/:keyword` returns Twitter API error 404:**

```
info: <-- GET /twitter/keyword/oled
error: Error in /twitter/keyword/oled: Error: Twitter API error: 404
info: --> GET /twitter/keyword/oled 503 904ms
```

X-side endpoint path appears to have changed (related: #19865, #20452 reported similar symptoms previously).

### Deployment information

Self-hosted

### Deployment information (for self-hosted)

[...]

PR #22137, opened 2026-05-30
fix(route/twitter): defensive fallback for missing profile_image_url (avoid /twitter/user/* 503)

Dinglingyun2022:

Defensive guard related to #22136 (does not fix the root cause; see "Scope" below).

```routes
/twitter/user/:id
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
    - [ ] Provided as ISO 8601 / UTC / 以 ISO 8601 / UTC 提供
- [ ] [Author](https://docs.rsshub.app/joinus/advanced/script-standard#%E5%88%9B%E5%BB%BA%E4%B8%80%E4%B8%AA%E4%BB%A3%E7%A0%81%E6%96%87%E4%BB%B6) in `namespace.ts` is included / `namespace.ts` 中含有 [作者](https://docs.rsshub.app/zh/joinus/advanced/script-standard#%E5%88%9B%E5%BB%BA%E4%B8%80%E4%B8%AA%E4%BB%A3%E7%A0%81%E6%96%87%E4%BB%B6)
- [X] Bug fix / 修复 bug

### Summary

`lib/routes/twitter/user.ts:91` calls `profileImageUrl.replace(...)` without optional chaining. When `api.getUser(id)` returns user info whose `profile_image_url` / `profile_image_url_https` fields are missing (observed when X changes the API contract — see #22136), `profileImageUrl` resolves to `undefined`, and the unguarded `.replace()` call throws `TypeError: Cannot read properties of undefined (reading 'replace')`, making the entire `/twitter/user/*` route return HTTP 503.

### Diff

```diff
-        image: profileImageUrl.replace(/_normal.jpg$/, '.jpg'),
+        image: profileImageUrl?.replace(/_normal.jpg$/, '.jpg') ?? 'https://abs.twimg.com/sticky/default_profile_images/default_profile_400x400.png',
```

One-line change: add optional chaining and fall back to X's official default profile image when the field is missing.

### Scope

- **Defensive only**: prevents a single missing field from collapsing the entire route into 503. Returns a feed with X's default avatar instead.
- **Does NOT fix the root cause**: the underlying X API contract change (which causes `profile_image_url` to be missing in the first place) is tracked in #22136 and requires updates to `gql-id-resolver` / query IDs / GraphQL features in `lib/routes/twitter/api/web-api/`.
- Does not affect the happy path: when `profileImageUrl` is present, behavior is unchanged.

### Why this is still valuable even after #22136 root cause lands

This is a generic defensive guard against future X API contract changes. X has historically renamed / removed fields on `UserByScreenName` multiple times. With this guard in place, future similar regressions degrade gracefully (default avatar) instead of crashing the entire route with a confusing `TypeError`.

### Symptom reference (from #22136)

```
info: <-- GET /twitter/user/dscassoc
info: twitter gql-id-resolver: fetching fresh query IDs from Twitter JS bundles
error: Error in /twitter/user/dscassoc: TypeError: Cannot read properties of undefined (reading 'replace')
info: --> GET /twitter/user/dscassoc 503 14s
```

### DIFF
--- a/lib/routes/twitter/user.ts
+++ b/lib/routes/twitter/user.ts
@@ -88,7 +88,7 @@ async function handler(ctx) {
     return {
         title: `Twitter @${userInfo?.name}`,
         link: `https://x.com/${userInfo?.screen_name}`,
-        image: profileImageUrl.replace(/_normal.jpg$/, '.jpg'),
+        image: profileImageUrl?.replace(/_normal.jpg$/, '.jpg') ?? 'https://abs.twimg.com/sticky/default_profile_images/default_profile_400x400.png',
         description: userInfo?.description,
         item:
             data &&


TonyRL:
Have you ever spent one second to check if https://x.com/dscassoc actually exists before wasting Mother Earth's precious energy feeding a worthless LLM that just spouts a wall of hallucinations?
image
A valid X account will **always** have `profile_image_url` even if they don't have profile image set. X gives them the common [default profile image](https://abs.twimg.com/sticky/default_profile_images/default_profile_normal.png) in the API response.

PR #22268, opened 2026-06-14
fix(route): repair u3c3 search and drop sticky promo rows

JaggerH:

Close #

```routes
/u3c3/search/狂飙
/u3c3/Video
```

- [x] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

This fixes the `u3c3` route, whose search has been silently broken — every query returned the same homepage listing (the recurring "always 國產原创" results).

**Three issues fixed:**

1. **Typo `keywoard`.** The handler destructured `const { type, keywoard, preview }` (misspelled), so `keyword` was always `undefined`, the `if (keyword)` search branch never ran, and every request fell through to fetching the homepage.

2. **Missing anti-scrape token.** u3c3 has since added a token to its search: its search box (`search21()` JS on the homepage) redirects to `/?search2=<token>&search=<kw>`. Hitting `/?search=<kw>` alone returns the homepage. The token rotates and is embedded in the page, so it is now scraped live from the homepage before building the search URL (with a graceful fallback to the old URL if it can't be found).

3. **Sticky promo rows.** u3c3 prepends fixed promo rows to every page — category-nav links (`href="/?type=…"`) and external ad links (`href="http://…"`, which surface as the `國產原创` entries with a malformed `guid`). These appear in search results too. The parser now keeps only rows whose detail link points to `/view`, which are the real torrent entries.

Verified against live u3c3: results are now query-specific (e.g. `狂飙` returns matching torrents; a title with no entries returns empty) instead of always returning the homepage, and the promo rows no longer appear.

### DIFF
--- a/lib/routes/u3c3/index.ts
+++ b/lib/routes/u3c3/index.ts
@@ -23,7 +23,7 @@ export const route: Route = {
 };
 
 async function handler(ctx) {
-    const { type, keywoard, preview } = ctx.req.param();
+    const { type, keyword, preview } = ctx.req.param();
     // should be:
     // undefined
     // U3C3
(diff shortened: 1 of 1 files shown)


lib/routes/u3c3/index.ts
```diff
@@ -52,22 +61,27 @@ async function handler(ctx) {
...
+            return [
+                {
+                    title: a.attr('title'),
+                    guid,
+                    link: guid,
+                    pubDate: item.find('td:nth-of-type(5)').text(),
+                    enclosure_url: item.find('td:nth-of-type(3) > a:nth-of-type(2)').attr('href'),
+                    enclosure_type: 'application/x-bittorrent',
+                },
+            ];
```
TonyRL: Unnecessary flatMap

JaggerH: Good catch — replaced the `flatMap` (which returned `[]`/`[obj]`) with a straightforward `.filter(...).map(...)`. Pushed in 5196476.

PR #22291, opened 2026-06-16
fix(route/btbtla): cover all seasons, tolerate dead rows, add lazy mode

JaggerH:

N/A

```routes
/btbtla/detail/雍正王朝
/btbtla/detail/雍正王朝/lazy
```

- [x] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

This improves the existing `btbtla/detail` route. It previously handled only a single exact-title match and a single quality tab, and one stale row could fail the whole feed.

- **Multi-season coverage.** The old route exact-matched the title (`a[title="${name}"]`), so a show listed with a season suffix (e.g. `上载新生 第三季`) returned nothing for the bare name. It now matches every search result whose title *contains* the query (deduped by href), covering all seasons/variants.
- **All quality tabs.** Previously only the active/selected quality tab was parsed; now every quality tab is included, tagged via `category`.
- **Fault tolerance.** btbtla leaves dead `/tdown/` pages that return 404. The previous per-row magnet fetch threw on those, failing the entire feed. Resolution now tolerates per-row failures and drops stale rows.
- **Bounded concurrency.** A popular show has 100+ download rows across seasons. Magnets are resolved with `p-map` (concurrency 5) + `cache.tryGet`, so refreshes are cheap (magnets are immutable) and we don't burst hundreds of requests at the site.
- **Optional `/lazy` mode.** Resolving every magnet up front is the dominant cost for large shows. `/btbtla/detail/:name/lazy` returns the `/tdown/` page links **without** resolving — for clients that resolve a single magnet on demand. The default stays **eager** so the feed carries real BT enclosures. It is a path param (not a query string) so it keys RSSHub's route cache distinctly — the cache key is the request path, query strings are ignored.

Season/quality info: quality goes into `category`; the season title and detail-page URL are attached under `_extra` (so they don't touch the standard rendered output but remain available to structured consumers reading the raw object / JSON Feed).

### DIFF
--- a/lib/routes/btbtla/detail.ts
+++ b/lib/routes/btbtla/detail.ts
@@ -1,14 +1,18 @@
 import { load } from 'cheerio'; // 类似 jQuery 的 API HTML 解析器
+import pMap from 'p-map';
 
-import type { Route } from '@/types';
+import type { DataItem, Route } from '@/types';
 import cache from '@/utils/cache';
 import ofetch from '@/utils/ofetch'; // 统一使用的请求库
 
 export const route: Route = {
-    path: '/detail/:name',
+    path: '/detail/:name/:lazy?',
     categories: ['multimedia'],
     example: '/btbtla/detail/雍正王朝',
-    parameters: { name: '电影 | 电视剧名称' },
+    parameters: {
+        name: '电影 | 电视剧名称',
+        lazy: 'Pass `lazy` to skip up-front magnet resolution (fast; no BT enclosure)',
+    },
     features: {
         requireConfig: false,
         requirePuppeteer: false,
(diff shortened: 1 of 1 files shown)


lib/routes/btbtla/detail.ts
```diff
@@ -20,72 +24,153 @@ export const route: Route = {
...
+    const seen = new Set<string>();
+    const out: Array<{ href: string; title: string }> = [];
+    $('.module-items .module-item-titlebox a[href^="/detail/"]').each((_, el) => {
+        const href = $(el).attr('href');
+        const title = $(el).attr('title') ?? '';
+        if (!href || seen.has(href) || !title.includes(name)) {
+            return;
+        }
+        seen.add(href);
+        out.push({ href, title });
```
TonyRL, line 160: Please provide the page URL that the search result returns duplicated href that requires using deduplication.

lib/routes/btbtla/detail.ts
```diff
@@ -20,72 +24,153 @@ export const route: Route = {
...
+            .each((_, info) => {
+                const a = $(info).find('a.module-row-text');
+                const href = a.attr('href');
+                if (!href) {
+                    return;
+                }
+                const title = $(info).find('.module-row-title h4').text().trim() || a.attr('title') || '';
+                rows.push({ title, downPage: ROOT_URL + href, quality, seasonTitle, seasonUrl });
+            });
+    });
```
TonyRL, line 140: Use [`Array.prototype.map`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/map) instead of [`Array.prototype.push`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/push)

lib/routes/btbtla/detail.ts
```diff
@@ -20,72 +24,153 @@ export const route: Route = {
...
-            };
-        })
-    );
-
-    const moduleTitle = 'BTBTLA | ' + $('.page-title').text();
+                items.push(buildItem(row, magnet));
+            },
+            { concurrency: CONCURRENCY }
+        );
+    }
```
TonyRL, line 88: Use [`Array.prototype.map`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/map) instead of [`Array.prototype.push`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/push)

PR #22464, opened 2026-07-07
feat(route): add route for 浙江大学数学科学院

Alex222222222222:

```routes
/zju/math/0
/zju/math/1
/zju/math/2
/zju/math/3
/zju/math/4
/zju/math/5
/zju/math/6
```

- [x] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

### DIFF
--- a/lib/routes/zju/math/index.ts
+++ b/lib/routes/zju/math/index.ts
@@ -61,8 +61,7 @@ async function enrichNewsItemWithDetails(item: DataItem, refererUrl: string): Pr
             const $ = load(response.data);
             const description = $('.wp_articlecontent').html();
             const infoText = $('.item_info').text();
-            const author = infoText.match(/来源：([\s\S]*?)发布时间：/)?.[1]?.trim();
-            const pubDate = infoText.match(/发布时间：(\d{4}-\d{2}-\d{2})/)?.[1];
+            const [, author, pubDate] = infoText.match(/来源：([\s\S]*?)发布时间：(\d{4}-\d{2}-\d{2})/) ?? [];
 
             if (description) {
                 item.description = description;
--- a/lib/routes/zju/math/index.ts
+++ b/lib/routes/zju/math/index.ts
@@ -1,0 +1,125 @@
+import { load } from 'cheerio';
+
+import type { DataItem, Route } from '@/types';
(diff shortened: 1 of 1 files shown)


lib/routes/zju/math/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+async function fetchNewsItemsByCategory(categoryId: string): Promise<DataItem[]> {
+    const response = await got({
+        method: 'get',
+        url: new URL(categoryId, base).href,
+    });
+
+    const $ = load(response.data);
+
+    return $('.news_list li')
```
TonyRL: Narrow down the selector to select the news only. It is picking up page navigation now.

lib/routes/zju/math/index.ts
```diff
@@ -0,0 +1,126 @@
...
+                headers: {
+                    Referer: refererUrl,
+                },
+            });
+
+            const $ = load(response.data);
+            const description = $('.wp_articlecontent').html();
+            const infoText = $('.item_info').text();
+            const author = infoText.match(/来源：([\s\S]*?)发布时间：/)?.[1]?.trim();
+            const pubDate = infoText.match(/发布时间：(\d{4}-\d{2}-\d{2})/)?.[1];
```
TonyRL: Use one regex with two matching groups instead of matching it twice.

lib/routes/zju/math/index.ts
```diff
@@ -0,0 +1,126 @@
...
+async function enrichNewsItemWithDetails(item: DataItem, refererUrl: string): Promise<DataItem> {
+    if (!item.link) {
+        return item;
+    }
+
+    return await cache.tryGet(item.link, async () => {
+        try {
+            const response = await got({
+                method: 'get',
+                url: item.link,
```
TonyRL: The article list does indicate which content is available only on the intranet. Use that information instead of blindly visiting every link.

Alex222222222222: Added intranet accessibility check based on the article list. There are some entries which is not indicated to be accessible from intranet only, but seems to redirect to ZJU webvpn or ZJU auth page, the current code still visit them.

PR #22497, opened 2026-07-09
feat(route): add Diari d'Andorra news

lisyer:
Add RSS route for [Diari d'Andorra](https://www.ara.ad/).

N/A

```routes
/ara
```

- [x] New Route / 新的路由
    - [x] Follows [Script Standard](https://docs.rsshub.app/joinus/advanced/script-standard)
- [ ] Anti-bot or rate limit
- [x] Date and time
    - [x] Parsed
    - [x] Correct time zone
- [ ] New package added
- [ ] `Puppeteer`

## Note / 说明

- WordPress REST API: `https://www.ara.ad/wp-json/wp/v2/posts`
- Route: `/ara`

### DIFF
--- a/lib/routes/ara/index.ts
+++ b/lib/routes/ara/index.ts
@@ -33,14 +33,13 @@ async function handler(ctx) {
     });
 
     const items = posts.map((item) => {
-        const $ = load(item.content?.rendered ?? '');
         const media = item._embedded?.['wp:featuredmedia']?.[0];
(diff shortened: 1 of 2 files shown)


TonyRL:
The proposed route duplicates content already available in the official RSS feeds and does not provide any unique value. Furthermore, the official RSS feeds are always updated with the latest content, while the proposed routes will always lag behind due to RSSHub internal caching.

PR #22534, opened 2026-07-09
feat(route): add Radio Okapi news

lisyer:
Add RSS route for [Radio Okapi](https://www.radiookapi.net/).

N/A

```routes
/radiookapi
```

- [x] New Route
    - [x] Follows Script Standard
- [ ] Anti-bot or rate limit
- [x] Date and time
    - [x] Parsed
    - [x] Correct time zone
- [ ] New package added
- [ ] Puppeteer

## Note / 说明

- Official feed: `https://www.radiookapi.net/rss.xml` (normalized via RSSHub)
- Route: `/radiookapi`

### DIFF
--- a/lib/routes/radiookapi/index.ts
+++ b/lib/routes/radiookapi/index.ts
@@ -1,0 +1,43 @@
+import { load } from 'cheerio';
+
+import type { Route } from '@/types';
+import { parseDate } from '@/utils/parse-date';
+import parser from '@/utils/rss-parser';
+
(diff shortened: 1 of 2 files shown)


TonyRL:
The proposed route duplicates content already available in the official RSS feeds and does not provide any unique value. Furthermore, the official RSS feeds are always updated with the latest content, while the proposed routes will always lag behind due to RSSHub internal caching.

PR #22708, opened 2026-07-13
fix(route/xueqiu): serialize show.json requests and skip caching transient failures

sirius60111:

NOROUTE

```routes
/xueqiu/user/8152922548
```

- [ ] New Route / 新的路由
- [ ] `Puppeteer`

## Note / 说明

Fix for the existing `/xueqiu/user/:id/:type?` route.

Concurrent `show.json` calls (via `Promise.all` on up to 20 items) triggered Xueqiu's rate limiting on `statuses/show.json`. Once the limit was hit, all subsequent calls in the same request were rejected. The `cache.tryGet` wrapper then cached the failed results containing empty descriptions, making the issue persist even after the rate limit cooled.

修复已有的 `/xueqiu/user` 路由。原实现用 `Promise.all` 并发请求最多 20 个 `show.json`，在雪球 `statuses/show.json` 接口上触发频率限流，同一请求内后续调用全部被拒。`cache.tryGet` 将失败结果（空 description）缓存，导致问题持续。

**Changes / 改动**

- **Serialize `show.json` calls**: replace `Promise.all(data.map(...))` with a `for...of` loop, making one `show.json` request at a time instead of firing up to 20 concurrently. / 串行请求: `Promise.all` → `for...of`，逐个请求而非 20 并发。
- **Skip caching transient failures**: when `show.json` fails with a rate limit or WAF error (no `error_code` in the response body), re-throw the error so `cache.tryGet` does not persist the result. The next request will retry. / 瞬态失败不进缓存: 限流/WAF 错误时抛异常，`cache.tryGet` 不存；下次请求重试。
- **Cache permanent failures**: when `show.json` returns a business error (e.g. deleted post with `error_code`), cache the fallback result as before. / 永久失败(已删帖等，有 `error_code`)仍缓存。
- An outer `try/catch` around each `cache.tryGet` call provides a fallback item from the list data when a transient failure occurs, keeping the feed valid while individual items retry on the next poll. / 外层 try/catch 在瞬态失败时提供兜底 item，item 在下次轮询时自动恢复。

### DIFF
--- a/lib/routes/xueqiu/user.ts
+++ b/lib/routes/xueqiu/user.ts
@@ -1,4 +1,5 @@
 import sanitizeHtml from 'sanitize-html';
+import pMap from 'p-map';
 
 import { parseToken } from '@/routes/xueqiu/cookies';
 import type { Route } from '@/types';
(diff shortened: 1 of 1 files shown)


TonyRL (commented):
Use p-map to limit the concurrent requests instead of resolving them one by one.

sirius60111:
@TonyRL Switched to p-map with concurrency limit of 3. PTAL, thanks!

PR #22720, opened 2026-07-14
fix(dongqiudi): adapt routes to new __NUXT__ data structure

wha7ev9r:

Close #

```routes
/dongqiudi/top_news/1
/dongqiudi/team_news/50001755
/dongqiudi/player_news/50000339
/dongqiudi/result/50001755
/dongqiudi/special/41
/dongqiudi/daily
```

- [ ] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

The dongqiudi website restructured its server-rendered `window.__NUXT__` data, breaking multiple dongqiudi routes. This PR adapts them to the new schema.

### `lib/routes/dongqiudi/utils.ts` (shared helpers)

- **`ProcessFeed`**: Use `teamInfo.name` / `detail.base_info` instead of `teamDetail.base_info` (field renames: `team_name` -> `name`, `team_logo` -> `logo`)
- **`ProcessFeed`**: Parse `show_time` as Unix seconds via `parseDate(article.show_time, 'X')` (was being parsed as milliseconds, producing 1970 dates)
- **`ProcessFeedType2`**: Read from `data[0].article` instead of `data[0].newData` (field renames: `writer` -> `author`, `body` -> `rawBody`)
- **`ProcessFeedType3`**: Removed the mobile parser because mobile pages can have empty `articleContent`, and no reliable case was found where mobile parsing succeeds while PC parsing fails.

### `lib/routes/dongqiudi/player-news.ts`

- Add missing `return` before `utils.ProcessFeed(...)` call. The handler was returning `undefined`, causing RSSHub to emit an empty feed.

### `lib/routes/dongqiudi/top-news.ts`

- Parse `show_time` as Unix seconds via `parseDate(item.show_time, 'X')` (same millisecond bug)

### `lib/routes/dongqiudi/result.ts`

- `teamScheduleData` has been removed from the team page `__NUXT__` data. Switch to the official API `api.dongqiudi.com/data/v1/team/schedule/{id}` to fetch schedule data.
- Fetch the last completed season's matches and filter by non-null scores.
- `pubDate` field renamed from `start_time` to `start_play` in the API response.
- Guard for teams with no prior season (return empty feed instead of throwing).

### `lib/routes/dongqiudi/special.ts`

- Use PC-end `ProcessFeedType2` for article content parsing. If an old article page returns `404` or another fetch-layer error, keep the list item instead of failing the whole feed.
- Removed the mobile `ProcessFeedType3` fallback because mobile pages can have empty `articleContent`, and no reliable case was found where mobile parsing succeeds while PC parsing fails.
- Removed unused `mobileLink` field.

### `lib/routes/dongqiudi/daily.ts`

- Add missing `return` before `ctx.set('redirect', ...)` call.

### Verification

All routes verified to return valid RSS via `pnpm dev`:

| Route | Test URL | Result |
|---|---|---|
| `top_news` | `/dongqiudi/top_news/1` | RSS with titles, descriptions, authors, pubDates |
| `team_news` | `/dongqiudi/team_news/50001755` | RSS for 皇家马德里 with team logo |
| `player_news` | `/dongqiudi/player_news/50000339` | RSS for 莫德里奇 with player photo |
| `result` | `/dongqiudi/result/50001755` | RSS with completed matches |
| `special` | `/dongqiudi/special/41` | RSS for 新闻大爆炸 using PC article data and skipping unavailable old article pages |
| `daily` | `/dongqiudi/daily` | Redirects to `/dongqiudi/special/48` |

### DIFF
--- a/lib/routes/dongqiudi/special.ts
+++ b/lib/routes/dongqiudi/special.ts
@@ -36,17 +36,8 @@ async function handler(ctx) {
     const out = await Promise.all(
         list.map((item) =>
             cache.tryGet(item.link, async () => {
-                try {
-                    const { data: response } = await got(item.link);
-                    const success = utils.ProcessFeedType2(item, response);
-                    if (!success) {
-                        throw new Error('No article data');
-                    }
-                } catch {
-                    const { data: mobileResponse } = await got(`https://m.dongqiudi.com/article/${item.link.match(/\d+/)[0]}.html`);
-                    utils.ProcessFeedType3(item, mobileResponse);
-                }
-
+                const { data: response } = await got(item.link);
+                utils.ProcessFeedType2(item, response);
                 return item;
             })
         )
--- a/lib/routes/dongqiudi/special.ts
+++ b/lib/routes/dongqiudi/special.ts
@@ -9,7 +9,9 @@ export const route: Route = {
     path: '/special/:id',
     categories: ['sport'],
     example: '/dongqiudi/special/41',
-    parameters: { id: '专题 id, 可自行通过 https://www.dongqiudi.com/special/+数字匹配' },
+    parameters: {
+        id: '专题 id, 可自行通过 https://www.dongqiudi.com/special/+数字匹配',
+    },
     radar: [
         {
             source: ['www.dongqiudi.com/special/:id'],
@@ -36,8 +38,15 @@ async function handler(ctx) {
     const out = await Promise.all(
         list.map((item) =>
             cache.tryGet(item.link, async () => {
-                const { data: response } = await got(item.link);
-                utils.ProcessFeedType2(item, response);
+                try {
+                    const { data: response } = await got(item.link);
+                    utils.ProcessFeedType2(item, response);
+                } catch (error) {
+                    if (!(error instanceof Error) || !['HTTPError', 'RequestError', 'FetchError'].includes(error.name)) {
+                        throw error;
+                    }
+                    // Keep the list item when the article page is gone or temporarily unavailable.
+                }
                 return item;
             })
         )
--- a/lib/routes/dongqiudi/special.ts
+++ b/lib/routes/dongqiudi/special.ts
@@ -30,16 +30,22 @@ async function handler(ctx) {
     const list = response.data.map((item) => ({
         title: item.title,
         link: `https://www.dongqiudi.com/articles/${item.aid}.html`,
-        mobileLink: `https://m.dongqiudi.com/article/${item.aid}.html`,
         pubDate: parseDate(item.show_time, 'X'),
     }));
 
     const out = await Promise.all(
         list.map((item) =>
             cache.tryGet(item.link, async () => {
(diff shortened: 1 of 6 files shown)


lib/routes/dongqiudi/special.ts
```diff
@@ -30,16 +30,22 @@ async function handler(ctx) {
...
+                try {
+                    const { data: response } = await got(item.link);
+                    const success = utils.ProcessFeedType2(item, response);
+                    if (!success) {
+                        throw new Error('No article data');
+                    }
+                } catch {
+                    const { data: mobileResponse } = await got(`https://m.dongqiudi.com/article/${item.link.match(/\d+/)[0]}.html`);
+                    utils.ProcessFeedType3(item, mobileResponse);
+                }
```
TonyRL, line 49: > Old mobile articles have empty `articleContent` on the mobile page.

If so, what's the point of using `ProcessFeedType3` in `catch`? Could you provide a URL that doesn't work with `ProcessFeedType2` but works with `ProcessFeedType3`?

lib/routes/dongqiudi/utils.ts
```diff
@@ -123,38 +125,39 @@ const ProcessFeedType2 = (item, response) => {
...
+    const initialState = JSON.parse(match[1]);
 
     // filter out undefined item
     if (!initialState) {
         return;
     }
 
-    if (Object.keys(initialState.articleContent).length) {
-        const data = Object.values(initialState.articleContent)[0];
+    if (initialState?.articleContent && Object.keys(initialState.articleContent).length) {
```
TonyRL: What's the point of optional chaining when there's https://github.com/DIYgod/RSSHub/blob/6f90d46003c043935c211134f711a66169cb3dc9/lib/routes/dongqiudi/utils.ts#L155-L157

wha7ev9r:
@TonyRL Thanks for the review.

I checked this again and couldn't find a reliable URL where `ProcessFeedType2` fails but `ProcessFeedType3` succeeds. Since old mobile pages may have empty `articleContent`, the mobile fallback does not seem useful here.

I removed the fallback in `special.ts` and deleted `ProcessFeedType3` from `utils.ts`, which also resolves the redundant optional chaining issue. The PR description has also been updated accordingly.

TonyRL (commented):
https://github.com/DIYgod/RSSHub/blob/ac84d7bac915147b7ec68bc95f2a8d510a99e4fa/lib/routes/dongqiudi/utils.ts#L132 https://github.com/DIYgod/RSSHub/blob/ac84d7bac915147b7ec68bc95f2a8d510a99e4fa/lib/routes/dongqiudi/utils.ts#L140
Why `ProcessFeedType2` is changed to return a bool?

wha7ev9r:
You're right — the bool return was a leftover from the previous if (!success) throw pattern in special.ts. Since no caller uses the return value anymore, I've reverted it to the original void return. @TonyRL  Please review it again ,thank!

PR #22770, opened 2026-07-20
feat(route): add Suzhou Museum Special Exhibit route

magazian:

Close #

```routes
/szmuseum/temporary
```

- [x] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

### DIFF
--- a/lib/routes/szmuseum/temporary.tsx
+++ b/lib/routes/szmuseum/temporary.tsx
@@ -1,13 +1,40 @@
-import tls from 'node:tls';
-
 import { load } from 'cheerio';
 import dayjs from 'dayjs';
 import { renderToString } from 'hono/jsx/dom/server';
+import { Agent, buildConnector } from 'undici';
 
 import type { Route } from '@/types';
+import ofetch from '@/utils/ofetch';
 import { parseDate } from '@/utils/parse-date';
 
 import { namespace } from './namespace';
+
+// szmuseum.com responds with a malformed `Cache Control: no-cache` header
+// (space instead of hyphen), which makes Node.js's strict HTTP parser throw.
+// Patch the bytes on the socket before the parser sees them.
+const connect = buildConnector({});
+const dispatcher = new Agent({
+    connect(opts, cb) {
+        connect(opts, (err, socket) => {
+            if (err || !socket) {
+                cb(err, null);
+                return;
+            }
+            const read = socket.read.bind(socket);
+            socket.read = (...args) => {
+                const chunk = read(...args);
+                if (Buffer.isBuffer(chunk)) {
+                    const index = chunk.indexOf('\r\nCache Control:');
+                    if (index !== -1) {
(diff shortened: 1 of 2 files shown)


magazian:
> ## Auto Review
> * **[API and Requesting / Script Standard]** `lib/routes/szmuseum/temporary.tsx`: The route fetches the page with a hand-rolled `node:tls` socket and manual HTTP parsing to work around a malformed `Cache-Control: no-cache ` (trailing space) header. This bypasses RSSHub's request layer entirely — no `config.trueUA`, no proxy support, and the raw parser doesn't handle chunked/gzipped transfer encodings, so it can silently corrupt the body. The strict trailing-whitespace header validation that breaks native `fetch`/`ofetch` (undici) does **not** affect `got`, which is built on `node:http` and tolerates such headers. Fix: replace the `tls.connect` block with the standard util, e.g. `import got from '@/utils/got';` then `const { data: content } = await got('https://www.szmuseum.com/Exhibition/Temporary');` and feed `content` into `load()`.
> 
> _Other items (the `_extra` field, dates/location in `description`, JSX rendering, radar/namespace/example config) match the already-merged `lnmuseum/temporary.tsx` pattern and look fine._

I tried `got` or `insecureHTTPParser`, it will still get error message: `error: Error in /szmuseum/temporary: FetchError: [GET] "https://www.szmuseum.com/Exhibition/Temporary": <no response> fetch failed (Response does not match the HTTP/1.1 protocol (Invalid header token))` since a malformed `Cache Control: no-cache`  is used in header (sorry I put the wrong message in the code comments). In this case, if `node: tls` is not allowed, do I need to use `Puppeteer` instead?

lib/routes/szmuseum/temporary.tsx
```diff
@@ -0,0 +1,139 @@
...
+            socket.on('data', (chunk) => {
+                data = Buffer.concat([data, Buffer.from(chunk)]);
+            });
+            socket.on('end', () => {
+                const str = data.toString('utf-8');
+                const bodyIndex = str.indexOf('\r\n\r\n');
+                resolve(str.slice(bodyIndex + 4));
+            });
+            socket.on('error', reject);
+        });
```
TonyRL: Override the default dispatcher with a custom one that patch the problematic header.
```ts
import { Agent, buildConnector } from 'undici';
const connect = buildConnector({});
const dispatcher = new Agent({
    connect(opts, cb) {
        connect(opts, (err, socket) => {
            if (err || !socket) {
                cb(err, null);
                return;
            }
            const read = socket.read.bind(socket);
            socket.read = (...args) => {
                const chunk = read(...args);
                if (Buffer.isBuffer(chunk)) {
                    const index = chunk.indexOf('\r\nCache Control:');
                    if (index !== -1) {
                        chunk.write('-', index + 7);
                    }
                }
                return chunk;
            };
            cb(null, socket);
        });
    },
});
```

PR #22804, opened 2026-07-23
fix(route/c114): parse article publication date

defp:

N/A

```routes
/c114/roll
```

- [ ] New Route / 新的路由
- [x] Date and time parsed with the correct time zone / 日期和时区已正确解析
- [ ] Puppeteer

## Note / 说明

- Parse publication dates from the current `.article_top .time` selector.
- Retain the legacy `div.r_time` selector as a fallback.
- Checks: ESLint, oxfmt, and `git diff --check`.

### DIFF
--- a/lib/routes/c114/roll.ts
+++ b/lib/routes/c114/roll.ts
@@ -51,7 +51,7 @@ export const handler = async (ctx) => {
 
                 item.title = title;
                 item.description = description;
-                item.pubDate = timezone(parseDate($$('div.r_time').text(), 'YYYY/M/D HH:mm'), 8);
+                item.pubDate = timezone(parseDate($$('.article_top .time, div.r_time').first().text().trim(), 'YYYY/M/D HH:mm'), 8);
                 item.author = $$('div.author').first().text().trim();
                 item.content = {
                     html: description,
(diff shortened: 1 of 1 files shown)


lib/routes/c114/roll.ts
```diff
@@ -51,7 +51,7 @@ export const handler = async (ctx) => {
 
                 item.title = title;
                 item.description = description;
-                item.pubDate = timezone(parseDate($$('div.r_time').text(), 'YYYY/M/D HH:mm'), 8);
+                item.pubDate = timezone(parseDate($$('.article_top .time, div.r_time').first().text().trim(), 'YYYY/M/D HH:mm'), 8);
```
TonyRL: Could you provide the site URL where the `div.r_time` is still being used for pubDate? Also, could you share the site URL where you found multiple `.article_top .time` elements that require using `first()` to select the correct element and contain leading or trailing spaces that require using `trim()`?

defp: https://www.c114.com.cn/news/103/a1314430.html 
image
.article_top .time

TonyRL: Yeah, I can see exactly one `.article_top .time` element with content that doesn't come with leading or trailing spaces that can be `trim()` from the link you provided.

Please read the question again.

> Could you share the site URL where you found **<ins>multiple</ins>** `.article_top .time` elements that require using `first()` to select the correct element and **<ins>contain leading or trailing spaces</ins>** that require using a `trim()`?

defp: del unused function

PR #22809, opened 2026-07-23
feat(route): add china silk museum exhibit route

magazian:

Close #

```routes
/chinasilkmuseum/zz
```

- [x] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

### DIFF
--- a/lib/routes/chinasilkmuseum/zz.tsx
+++ b/lib/routes/chinasilkmuseum/zz.tsx
@@ -1,0 +1,120 @@
+import { load } from 'cheerio';
+import dayjs from 'dayjs';
+import { renderToString } from 'hono/jsx/dom/server';
+
+import type { Data, DataItem, Route } from '@/types';
+import cache from '@/utils/cache';
+import ofetch from '@/utils/ofetch';
+import { parseDate } from '@/utils/parse-date';
+
+import { namespace } from './namespace';
+
+// Convert YYYY年M月D日 to YYYY-MM-DD
+const fmtExhibitionDate = (raw: string | undefined) => {
+    if (!raw) {
+        return;
+    }
+    const d = dayjs(raw.trim(), 'YYYY年M月D日');
+    return d.format('YYYY-MM-DD');
+};
+
+// used for 2026年6月24日 - 2026年9月1日
+const parseExhibitionDuration = (fullDuration: string): { startDate: string | undefined; endDate: string | undefined } => {
+    const [startRaw, endRaw] = fullDuration.split(' - ', 2);
+    return { startDate: fmtExhibitionDate(startRaw), endDate: fmtExhibitionDate(endRaw) };
+};
+
+export const route: Route = {
+    path: '/zz',
+    categories: ['travel'],
+    example: '/chinasilkmuseum/zz',
+    name: 'Exhibition',
+    maintainers: ['magazian'],
+    radar: [
+        {
+            source: ['www.chinasilkmuseum.com/zz/list_17.aspx'],
+            target: '/zz',
+        },
+    ],
+    handler: async () => {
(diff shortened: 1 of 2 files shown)


lib/routes/chinasilkmuseum/zz.tsx
```diff
@@ -0,0 +1,120 @@
...
+        const items = $('div.about_body div.show_info ul.ul li')
+            .toArray()
+            .map((el) => {
+                const $el = $(el);
+                const $titleLink = $el.find('div.show_text h3.h3 a');
+                const title = $titleLink.text();
+                const rawHref = $titleLink.attr('href') ?? '';
+                const link = new URL(rawHref, baseUrl).href;
+
+                const $img = $el.find('a > img').first();
```
TonyRL: Have you found a second `a > img` element within each `div.about_body div.show_info ul.ul li` element that require using `first()` to select the correct element?

magazian:
> ## Auto Review
> ### Route Metadata and Docs
> * **[Radar source]** `lib/routes/chinasilkmuseum/zz.tsx:33` — `source: ['chinasilkmuseum.com/zz/list_17.aspx']` has no subdomain, but the namespace `url` is `www.chinasilkmuseum.com`. `scripts/workflow/build-routes.ts` buckets radar rules by `subdomain` parsed from `source[0]`, so this rule is registered under `.` and will not fire on `www.chinasilkmuseum.com`. Fix: `source: ['www.chinasilkmuseum.com/zz/list_17.aspx']`.
> 
> ### Data Handling and Feed Quality
> * **[description content]** `lib/routes/chinasilkmuseum/zz.tsx:69-90` — the `description` is built entirely from metadata (`地点` / `开展` / `闭展` / `原始展期`) and re-states the start date that is already exposed as `pubDate`; the actual exhibition introduction from the detail page is never included. Fix: render the detail page's main content (e.g. the exhibition intro block from `$d`) as `description`, keep the poster ``, and drop the `开展`/`闭展`/`原始展期` lines since the date belongs to `pubDate`.
> * **[invalid date leaks into the feed]** `lib/routes/chinasilkmuseum/zz.tsx:11-18` — `fmtExhibitionDate` returns `d.format('YYYY-MM-DD')` unconditionally, so a non-matching duration string yields the literal string `'Invalid Date'`, which is then rendered in the description and passed to `parseDate()`, producing an invalid `pubDate` instead of no date. Fix: `return d.isValid() ? d.format('YYYY-MM-DD') : undefined;`.
> * **[pubDate timezone]** `lib/routes/chinasilkmuseum/zz.tsx:57` — `parseDate('YYYY-MM-DD')` is interpreted in the runtime's local/UTC zone, so the date is off by 8 hours for this CST source. Fix: parse the raw value directly with the timezone helper and drop the `dayjs` round-trip, e.g. `timezone(parseDate(startRaw, 'YYYY年M月D日'), +8)` (`import timezone from '@/utils/timezone'`).
> 
> ### Code Style and Maintainability
> * **[dead fallback]** `lib/routes/chinasilkmuseum/zz.tsx:74` — `location ?? '参考详情'` can never take the fallback: `$d(...).first().text().replace(...).trim()` always returns a string (`''` when the element is missing). Fix: use `location || '参考详情'` (or omit the line entirely if the metadata block is removed per the `description` item above).

* **[Radar source]** Updated accordingly
* **[description content]** Would like to keep the original logic, since `地点` / `开展` / `闭展` / `原始展期` is importance information for the exhibition. However the `pubdate` is never defined from the website, use `startDate` as `purDate` and it would not be used directly in the `description` according to my understanding
* **[invalid date leaks into the feed]** Updated accordingly
* **[pubDate timezone]** As metioned above, `purDate` is not accurate, so the timezone doesn't matter, would like to keep as it is.
* **[dead fallback]** Updated accordingly

PR #22852, opened 2026-07-27
feat(core): normalize fallback logic for content.text and content.html in Atom, RSS and JSON Feed

gosxrgxx:

Close #

```routes
NOROUTE
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

#### Atom

Atom, like JSON Feed, [supports](https://datatracker.ietf.org/doc/html/rfc4287#section-4.1.2:~:text=On%20the%20atom%3Acontent%20element%2C%20the%20value%20of%20the%20%22type%22%20attribute%20MAY%20be%0A%20%20%20one%20of%20%22text%22%2C%20%22html%22) `content.text` (via the `type="text"` attribute), but [unlike](https://www.jsonfeed.org/version/1.1/#items:~:text=content%5Fhtml%20and%20content%5Ftext%20are%20each%20optional%20strings%20%E2%80%94%20but%20one%20or%20both%20must%20be%20present%2E) JSON Feed, the Atom element (`atom:entry`) [must not](https://datatracker.ietf.org/doc/html/rfc4287#section-4.1.2:~:text=atom%3Aentry%20elements%20MUST%20NOT%20contain%20more%20than%20one%20atom%3Acontent%20element) contain more than one `content` element.

Therefore, a condition was added for Atom where `content` with the `type="text"` attribute acts as a fallback if `content.html` is missing.

If `content` is empty, it [must](https://datatracker.ietf.org/doc/html/rfc4287#section-4.1.3.2) contain a link (`item.link`) to the content in the `src` attribute.

#### RSS

Also, `content.html` has been added as a fallback for the `description` in the middleware. This removes the need to include a `description` when creating new routes, allowing the use of clearer options like `content.html`. At the same time, this maintains backward compatibility with old routes.

#### JSON Feed

`item.title` must be rendered in `content_html` only if `content.text` is not provided, to prevent getting the text content in `content_text` and the title in `content_html`. Also, since `item.title` is text content, it must be a fallback for `content_text` instead of `content_html`.

### DIFF
--- a/lib/views/json.ts
+++ b/lib/views/json.ts
@@ -20,8 +20,8 @@ const json = (data: Data) => {
             url: item.link,
             title: item.title,
             // content_html and content_text are each optional strings — but one or both must be present
-            content_html: (item.content && item.content.html) || item.description || item.title,
-            content_text: item.content?.text ?? undefined,
+            content_html: item.content?.html || item.description,
+            content_text: item.content?.text || (item.content?.html || item.description ? undefined : item.title),
             summary: item.summary,
             image: item.image || item.itunes_item_image,
             banner_image: item.banner,
--- a/lib/views/json.ts
+++ b/lib/views/json.ts
@@ -20,8 +20,8 @@ const json = (data: Data) => {
             url: item.link,
             title: item.title,
             // content_html and content_text are each optional strings — but one or both must be present
-            content_html: item.description,
-            content_text: item.content?.text || (item.description ? undefined : item.title),
+            content_html: item.description || item.content?.html,
+            content_text: item.content?.text || (item.description || item.content?.html ? undefined : item.title),
             summary: item.summary,
             image: item.image || item.itunes_item_image,
             banner_image: item.banner,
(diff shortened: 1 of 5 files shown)


lib/views/json.ts
```diff
@@ -20,8 +20,8 @@ const json = (data: Data) => {
             url: item.link,
             title: item.title,
             // content_html and content_text are each optional strings — but one or both must be present
-            content_html: (item.content && item.content.html) || item.description || item.title,
-            content_text: item.content?.text ?? undefined,
+            content_html: item.description || item.content?.html,
+            content_text: item.content?.text || (item.description || item.content?.html ? undefined : item.title),
```
TonyRL: With https://github.com/DIYgod/RSSHub/blob/7c067424c19094606ca588637ac86c4c0bd0dd3f/lib/middleware/parameter.ts#L89

`|| item.content?.html` is meaningless since every route's DataItem will go through `parameter` before reaching `views`.

PR #22906, opened 2026-08-02
feat(route/huggingface): support trending papers

HernandoR:

Close #

```routes
/huggingface/daily-papers
/huggingface/daily-papers/date
/huggingface/daily-papers/week
/huggingface/daily-papers/month
/huggingface/daily-papers/trending
/huggingface/daily-papers/trending/50
```

- [x] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Adds `trending` as a fourth cycle to the existing `huggingface/daily-papers` route, covering <https://huggingface.co/papers/trending>.

That page is server-rendered from the same `div[data-target="DailyPapers"]` props blob as the date/week/month listings, with the same item shape, so it slots into the existing handler rather than needing a new route file. The existing radar rule `huggingface.co/papers/:cycle` already resolved that URL to `/huggingface/daily-papers/trending`, which previously threw `Invalid cycle` — this makes the rule actually work.

Two small related changes:

- **Trending order is preserved.** The other cycles are re-sorted by upvotes; for trending the page order *is* the ranking, so it is passed through untouched (visible with `?sorted=false`).
- **Feed `title`/`link` now follow the cycle** (`Huggingface Daily/Weekly/Monthly/Trending Papers`, linking to the page actually fetched). Previously all four cycles produced a feed titled `Huggingface Daily Papers` pointing at `/papers`, which made them indistinguishable in a reader.

Verified locally against a dev server — all four cycles return 200 with items, the vote filter still applies (`/trending/50`), an unknown cycle still errors, and `/huggingface/daily-papers/trending?sorted=false` matches the site's trending order item-for-item.

### DIFF
--- a/lib/routes/huggingface/daily-papers.ts
+++ b/lib/routes/huggingface/daily-papers.ts
@@ -8,7 +8,7 @@ export const route: Route = {
     path: '/daily-papers/:cycle?/:voteFliter?',
     categories: ['programming'],
     example: '/huggingface/daily-papers/week/50',
-    parameters: { cycle: 'The publication cycle you want to follow. Choose from: date, week, month. Default: date', voteFliter: 'Filter papers by vote count.' },
+    parameters: { cycle: 'The publication cycle you want to follow. Choose from: date, week, month, trending. Default: date', voteFliter: 'Filter papers by vote count.' },
     features: {
         requireConfig: false,
         requirePuppeteer: false,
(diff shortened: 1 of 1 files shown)


lib/routes/huggingface/daily-papers.ts
```diff
@@ -77,13 +85,15 @@ async function handler(ctx) {
             pubDate: parseDate(item.publishedAt),
             author: item.paper.authors.map((author) => author.name).join(', '),
             upvotes: item.paper.upvotes,
-        }))
-        .toSorted((a, b) => b.upvotes - a.upvotes);
+        }));
+
+    // The trending page is already ranked by Huggingface, the cycle listings are not
+    const sortedItems = cycle === 'trending' ? items : items.toSorted((a, b) => b.upvotes - a.upvotes);
```
TonyRL, line 91: Please provide evidence showing where you found the trending page is ranked.

PR #22968, opened 2026-08-09
fix(route/futunn): render article pages via playwright to bypass WAF

wangjuns:

N/A — bug fix for existing routes.

```routes
/futunn/main
/futunn/topic/1267
```

- [ ] New Route / 新的路由
    - [x] Parsed / 可以解析
- [x] `Puppeteer`

### Problem

Futu's WAF now returns an obfuscated JavaScript shell to requests without a browser fingerprint (e.g. plain `got` requests). As a result, article bodies in the `/futunn/main` and `/futunn/topic` feeds have been **empty** — the pages fetched by `got` contain no `.origin_content` element, so `item.description` silently falls back to the abstract (or nothing).

### Changes

- Add `lib/routes/futunn/utils.ts` with:
  - `fetchArticleDetail(context, link)`: tries `got` first; if the response lacks `.origin_content` (WAF shell), falls back to rendering the page with **playwright** (via `@/utils/playwright`), then re-extracts the body.
  - `extractArticleInfo(content)`: shared body/category extraction previously duplicated in `main.ts` and `topic.ts`.
- Update `/futunn/main` and `/futunn/topic` to use the shared helper and mark `requirePuppeteer: true`.

### Verification

- `tsc --noEmit` passes; `oxlint` and `oxfmt` clean on `lib/routes/futunn/`.
- Locally ran the modified routes (`/futunn/main?limit=2`): article bodies now come through fully (e.g. the CPI article renders ~780 chars of body text instead of empty `<p></p>`).

### Notes

- The fallback only activates when `got` is blocked, so deployments without a browser keep current behavior (empty body) rather than failing.
- If playwright also fails (e.g. page itself returns 403), the entry degrades gracefully to the existing empty-description behavior.

### DIFF
--- a/lib/routes/futunn/utils.ts
+++ b/lib/routes/futunn/utils.ts
@@ -1,0 +1,72 @@
+import type { CheerioAPI } from 'cheerio';
+import { load } from 'cheerio';
+import type { BrowserContext } from 'patchright';
+
+import got from '@/utils/got';
+import logger from '@/utils/logger';
+
+/**
+ * 富途 WAF 会拦截无浏览器指纹的请求(如 got), 返回混淆 JS 壳页面,
+ * 其中不含 `.origin_content`, 导致条目正文为空。
+ * 此函数先尝试 got, 若检测到被 WAF 拦截, 则改用 playwright 渲染页面后重新提取。
+ */
+export async function fetchArticleDetail(context: BrowserContext, link: string): Promise<CheerioAPI> {
+    let html = '';
+
+    try {
+        const response = await got({ method: 'get', url: link });
+        html = response.data;
+
+        // got 拿到正常页面时直接返回
+        if (load(html)('.origin_content').length > 0) {
+            return load(html);
+        }
+    } catch {
+        // got 抛错(如 403), 继续走 playwright
+    }
+
+    logger.warn(`futunn: got 请求可能被 WAF 拦截, 改用 playwright 渲染: ${link}`);
+
+    try {
+        const page = await context.newPage();
+        try {
+            // 只加载文档与脚本, 跳过图片/样式/字体等资源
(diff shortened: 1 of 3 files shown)


lib/routes/futunn/utils.ts
TonyRL, line 1: Unnecessary playwright usage. This wafToken can be solved locally.

```ts
/**
 * Solve wafToken
 * pick a random integer number < 1e12 (and >= 10^max(Arr)) whose digits at 1-indexed positions `Arr[0]` and `Arr[1]` counted from the right both equal `Value`
 * hash the JWT with a rolling h = (132*h + charCode) mod 2^25 seeded at the FNV-1a basis 0x811c9dc5, hex-encoded
 * `${jwt}@${number}@${hash(jwt)}`
 */
const solveWafToken = (html: string) => {
    const jwt = // extract it from html
    const payload = Buffer.from(jwt.split('.', 2)[1], 'base64url').toString();
    const { Arr: arr, Value: value } = JSON.parse(JSON.parse(payload).args);

    let number;
    do {
        number = Math.floor(Math.random() * 1_000_000_000_000);
    } while (number < 10 ** Math.max(...Arr) || Arr.some((position) => Math.floor((number / 10 ** (position - 1)) % 10) !== Value));

    let hash = 0x81_1c_9d_c5;
    for (const char of jwt) {
        hash = (0x84 * hash + char.codePointAt(0)!) % 2 ** 25;
    }

    return `${jwt}@${number}@${hash.toString(16)}`;
};
```

`ofetch(link)`. If the HTML has `wafToken=`, `solveWafToken(html)` and `ofetch` again with `{ headers: { cookie: wafToken=result } }`.

TonyRL (changes requested):

PR #23024, opened 2026-08-13
fix(route/openai): adapt cookbook to new Astro-based site

LeKZzzz:

Close #

```routes
/openai/cookbook
```

- [ ] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

The OpenAI Cookbook site has migrated from `cookbook.openai.com` (Next.js + Tailwind CSS) to `developers.openai.com/cookbook` (Astro framework). All CSS selectors in the existing route no longer match, resulting in an empty feed.

**Changes:**

- Update URL to `developers.openai.com/cookbook`
- Extract article data from the `astro-island` component's serialized `props` JSON instead of parsing Tailwind CSS card selectors
- Add `decodeAstroValue()` to recursively decode the Astro serialization format (`[0, value]` scalar / `[1, [...]]` array)
- Filter out archived entries, sort by date descending
- Use `parseDate()` for proper Date objects
- Add `features` (`antiCrawler: true`), `parameters` (`limit`), update `radar` rules
- Add `User-Agent` header to bypass Vercel bot detection (403)
- Detail page selector updated from specific Tailwind classes to generic `article` element
- Add `limit` query parameter (default 30) to control the number of articles fetched

### DIFF
--- a/lib/routes/openai/cookbook.ts
+++ b/lib/routes/openai/cookbook.ts
@@ -1,78 +1,141 @@
+import type { Context } from 'hono';
 import { load } from 'cheerio';
 
+import { config } from '@/config';
 import type { DataItem, Route } from '@/types';
 import cache from '@/utils/cache';
-import logger from '@/utils/logger';
+import { parseDate } from '@/utils/parse-date';
 import ofetch from '@/utils/ofetch';
 
 export const route: Route = {
     path: '/cookbook',
     categories: ['programming'],
-    description:
-        'OpenAI Cookbook 提供了大量使用 OpenAI API 的实用指南和示例代码，涵盖了从基础到高级的各种主题，包括 GPT 模型、嵌入、函数调用、微调等。这里汇集了最新的 API 功能介绍和流行的应用案例，是开发者学习和应用 OpenAI 技术的宝贵资源。',
-    maintainers: ['liyaozhong'],
+    example: '/openai/cookbook',
+    parameters: {
+        limit: '返回的文章数量，默认 30',
+    },
+    features: {
+        requireConfig: false,
+        requirePuppeteer: false,
+        antiCrawler: true,
+        supportBT: false,
+        supportPodcast: false,
+        supportScihub: false,
+    },
     radar: [
         {
-            source: ['cookbook.openai.com/'],
+            source: ['developers.openai.com/cookbook', 'cookbook.openai.com/'],
         },
     ],
-    url: 'cookbook.openai.com/',
-    handler,
-    example: '/openai/cookbook',
+    url: 'developers.openai.com/cookbook',
     name: 'Cookbook',
+    description: 'OpenAI Cookbook 提供了大量使用 OpenAI API 的实用指南和示例代码，涵盖了从基础到高级的各种主题，包括 GPT 模型、嵌入、函数调用、微调等。这里汇集了最新的 API 功能介绍和流行的应用案例，是开发者学习和应用 OpenAI 技术的宝贵资源。',
+    maintainers: ['liyaozhong'],
+    handler,
 };
 
-async function handler() {
-    const rootUrl = 'https://cookbook.openai.com';
-    const currentUrl = `${rootUrl}/`;
-
-    try {
-        const response = await ofetch(currentUrl);
-        const $ = load(response);
-
-        let items = $('[class="min-h-[90vh] mt-4"] .grid a')
-            .toArray()
-            .map((element): DataItem => {
-                const $element = $(element);
-                const $title = $element.find('div.font-semibold.text-sm.text-primary.line-clamp-1.overflow-ellipsis');
-                const $date = $element.find(String.raw`span.text-xs.text-muted-foreground.md\:w-24.text-end`);
-                const $author = $element.find('p:contains("OpenAI")');
-                const $tags = $element.find('span[style^="color:"]');
-
-                return {
-                    title: $title.text().trim(),
-                    link: `${rootUrl}/${$element.attr('href')}`,
-                    pubDate: $date.text().trim(),
-                    author: $author.text().replace('OpenAI', '').trim(),
-                    category: $tags.toArray().map((tag) => $(tag).text().trim()),
-                };
-            });
-
-        items = (
-            await Promise.all(
-                items.map((item) =>
-                    cache.tryGet(item.link!, async () => {
-                        try {
-                            const detailResponse = await ofetch(item.link!);
-                            const $ = load(detailResponse);
-
-                            item.description = $(String.raw`article.prose.prose-sm.sm\:prose-base.max-w-none.dark\:prose-invert`).html();
-                            return item;
-                        } catch {
-                            return { ...item, description: '' };
-                        }
-                    })
-                )
-            )
-        ).filter((item) => item?.description);
-
-        return {
-            title: 'OpenAI Cookbook',
-            link: currentUrl,
-            item: items,
-        };
-    } catch (error) {
-        logger.error(`处理 OpenAI Cookbook 请求时发生错误: ${error}`);
-        throw error;
+// 递归解码 Astro 组件 props 的序列化格式
+// [0, value] → 标量；[1, [...]] → 数组；普通对象递归解码每个属性
+function decodeAstroValue(val: unknown): unknown {
+    if (Array.isArray(val)) {
+        if (val.length === 2 && (val[0] === 0 || val[0] === 1)) {
+            if (val[0] === 0) {
+                return decodeAstroValue(val[1]);
+            }
+            if (Array.isArray(val[1])) {
+                return val[1].map(decodeAstroValue);
+            }
+        }
+        return val;
     }
+    if (val !== null && typeof val === 'object') {
+        const result: Record<string, unknown> = {};
+        for (const [key, value] of Object.entries(val)) {
+            result[key] = decodeAstroValue(value);
+        }
+        return result;
+    }
+    return val;
+}
+
+interface CookbookEntry {
+    title: string;
+    date: string;
+    slug: string;
+    internal_path: string;
+    description: string;
+    tags: string[];
+    authors?: { name: string }[];
+    archived: boolean;
+}
+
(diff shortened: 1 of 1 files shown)


lib/routes/openai/cookbook.ts
```diff
@@ -1,78 +1,142 @@
...
         },
     ],
-    url: 'cookbook.openai.com/',
-    handler,
-    example: '/openai/cookbook',
+    url: 'developers.openai.com/cookbook',
     name: 'Cookbook',
+    description:
+        'OpenAI Cookbook 提供了大量使用 OpenAI API 的实用指南和示例代码，涵盖了从基础到高级的各种主题，包括 GPT 模型、嵌入、函数调用、微调等。这里汇集了最新的 API 功能介绍和流行的应用案例，是开发者学习和应用 OpenAI 技术的宝贵资源。',
+    maintainers: ['liyaozhong'],
```
TonyRL: Why reorder `example`, `name`, `description`, `maintainers` and `handler`? Does the feed  work only when those properties are in this order?

lib/routes/openai/cookbook.ts
```diff
@@ -1,78 +1,142 @@
...
-            link: currentUrl,
-            item: items,
-        };
-    } catch (error) {
-        logger.error(`处理 OpenAI Cookbook 请求时发生错误: ${error}`);
-        throw error;
+    const rootUrl = 'https://developers.openai.com';
+    const currentUrl = `${rootUrl}/cookbook/`;
+
+    const response = await ofetch(currentUrl, {
```
TonyRL: Use the [redirected URL](https://github.com/DIYgod/RSSHub/actions/runs/34738525862/job/103674121966#step:14:22) instead.

lib/routes/openai/cookbook.ts
```diff
@@ -1,78 +1,142 @@
...
+        headers: { 'User-Agent': config.trueUA },
```
TonyRL: Why use this UA? Does the feed break when using the default UA, which is a randomised Chrome user agent on macOS, and work only when this fixed UA string is specified?

lib/routes/openai/cookbook.ts
```diff
@@ -1,78 +1,142 @@
...
+    });
+    const $ = load(response);
+
+    // Extract article data from the astro-island props attribute.
+    const propsRaw = $('astro-island[component-url*="CookbookHomepage"], astro-island[component-url*="CookbookSearchPage"]').first().attr('props');
```
TonyRL: Can you show me how the content of `astro-island[component-url*="CookbookSearchPage"]` look like?

TonyRL: Can you show me the page structure that contains multiple `astro-island[component-url*="CookbookHomepage"], astro-island[component-url*="CookbookSearchPage"]` elements that require using `first()` to select the correct one?

lib/routes/openai/cookbook.ts
```diff
@@ -1,78 +1,142 @@
...
+function decodeAstroValue(val: unknown): unknown {
+    if (Array.isArray(val)) {
+        if (val.length === 2 && (val[0] === 0 || val[0] === 1)) {
+            if (val[0] === 0) {
+                return decodeAstroValue(val[1]);
+            }
+            if (Array.isArray(val[1])) {
+                return val[1].map(decodeAstroValue);
+            }
+        }
```
TonyRL:
```diff
-        if (val.length === 2 && (val[0] === 0 || val[0] === 1)) {
-            if (val[0] === 0) {
-                return decodeAstroValue(val[1]);
-            }
-            if (Array.isArray(val[1])) {
-                return val[1].map(decodeAstroValue);
-            }
-        }
+        return val[0] === 0 ? decodeAstroValue(val[1]) : val[1].map((v) => decodeAstroValue(v));
```

lib/routes/openai/cookbook.ts
```diff
@@ -1,78 +1,142 @@
...
+                } else if (item.authors?.length) {
+                    dataItem.author = item.authors
+                        .map((a) => a.name)
+                        .filter(Boolean)
+                        .join(', ');
+                }
+
+                try {
+                    const detailResponse = await ofetch(dataItem.link!, {
+                        headers: { 'User-Agent': config.trueUA },
```
TonyRL: Why use this UA? Does the feed break when using the default UA, which is a randomised Chrome user agent on macOS, and work only when this fixed UA string is specified?

lib/routes/openai/cookbook.ts
```diff
@@ -1,78 +1,142 @@
...
+                    });
+                    const $detail = load(detailResponse);
+                    const content = $detail('article').first().html();
```
TonyRL: Can you provide an example article that has multiple `<article>` that require using `first()` to select the correct one?

lib/routes/openai/cookbook.ts
```diff
@@ -1,78 +1,142 @@
...
+                    if (content) {
+                        dataItem.description = content;
+                    }
+                } catch {
+                    // Keep the props description as a fallback.
+                }
```
TonyRL: Can you show me which article will emit a network error that require using try-catch?

TonyRL (changes requested):

lib/routes/openai/cookbook.ts
LeKZzzz: You're right, there is no functional reason for the reordering; the feed does not depend on the property order. It was incidental churn from editing the object while adding `features` and updating `url`/`radar`. I'll restore the original order from master and keep only the necessary field changes.

lib/routes/openai/cookbook.ts
LeKZzzz: Confirmed, and it's not a new behavior: `https://developers.openai.com/cookbook/` 308-redirects to `https://developers.openai.com/cookbook`, and the same redirect already appears in the CI log of the first submission on Aug 13 ([[run 31675349390](https://github.com/DIYgod/RSSHub/actions/runs/31675349390)](https://github.com/DIYgod/RSSHub/actions/runs/31675349390)):

```
debug: Outgoing request: GET https://developers.openai.com/cookbook/
http: Redirecting to https://developers.openai.com/cookbook for https://developers.openai.com/cookbook/
```

So it's been the site's behavior since the Astro migration, the route only worked because ofetch follows redirects. As you suggested, I'll use the final URL directly to avoid the extra round trip

lib/routes/openai/cookbook.ts
LeKZzzz: I can't demonstrate that this fixed UA is required: the Aug 13 CI run fetched successfully with `config.ua` , and the Sep 13 run fetched successfully with `config.trueUA`. The header was added while investigating a 403 I hit locally from Node; it later turned out to be unrelated to the UA (the same UA returns 200 through other clients — it was a local environment issue). Since I can't show that the default UA fails, the override is unjustified. I'll remove it and rely on RSSHub's default generated browser headers, with CI verifying the result

lib/routes/openai/cookbook.ts
LeKZzzz: That selector matched the real page structure at submission time: on Aug 13 the homepage island was named `CookbookSearchPage`. The evidence is the Aug 13 CI run ([[31675349390](https://github.com/DIYgod/RSSHub/actions/runs/31675349390)](https://github.com/DIYgod/RSSHub/actions/runs/31675349390)), which produced a feed with real articles using code that only matched that name, if the selector hadn't matched, `propsRaw` would be empty and the handler would throw, so no feed could be generated. Unfortunately I didn't keep an HTML snapshot from that day, so I can't paste the original markup.

OpenAI has since changed it: as of Sep 13 the component is named `CookbookHomepage`, which broke the selector again — that's why my second commit matches both names. The current structure is:

```
<astro-island component-url="/_astro/CookbookHomepage.react.CXeImWjD.js"
  props='{"entries":[1,[[0,{"internal_path":[0,"examples/agents_api/apps/sev_bot/readme"],
  "title":[0,"Build an SRE agent for incident response"],"date":[0,"2026-09-10"],
  "tags":[1,[[0,"agents"],[0,"agents-api"],[0,"sandbox"]]],
  "authorAffiliations":[1,[[0,"openai"]]]}],...]]}'>
```

LeKZzzz: There is no such structure.  `.first()` is unnecessary. I'll drop it and simplify to a single selector

lib/routes/openai/cookbook.ts
LeKZzzz: Agreed, the nested checks are needlessly verbose; I'll adopt your version.

lib/routes/openai/cookbook.ts
LeKZzzz: Same as the homepage case above: there's no evidence the fixed UA is needed. I'll remove this header as well and rely on the default headers.

lib/routes/openai/cookbook.ts
LeKZzzz: I checked the first 30 entries of the current list and every page has just that one. so I'll remove it

lib/routes/openai/cookbook.ts
LeKZzzz: I can't either, I checked the current first 30 detail pages and they all return 200. That try/catch was carried over from the pre-change implementation. The props currently carry no `description` field, so a failure would just produce an item with no content, and that degraded result would be cached by `cache.tryGet`. I'll remove the try/catch and let failures surface instead of caching a degraded result.

lib/routes/openai/cookbook.ts
```diff
@@ -2,77 +2,125 @@ import { load } from 'cheerio';
...
+                        .filter(Boolean)
+                        .join(', ');
+                }
+
+                const detailResponse = await ofetch(link);
+                const $detail = load(detailResponse);
+                const content = $detail('article').html();
+                if (content) {
+                    dataItem.description = content;
+                }
```
TonyRL:
```suggestion
                dataItem.description = content ?? dataItem.description;
```

PR #23082, opened 2026-08-24
fix(route/theverge): adapt to schema changes and harden article rendering

IvanWng97:

N/A

```routes
/theverge
/theverge/apple
/theverge/games
/theverge/reviews
```

- [ ] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

`/theverge` currently fails with `TypeError: Cannot read properties of undefined (reading 'map')`.

**Root cause.** The Verge renamed fields in its `__NEXT_DATA__` GraphQL payload: `CoreParagraphBlockType.tempContents` → `paragraphContents`, and stream excerpt entries changed from `{ contents: { html } }` to `{ paragraphContents: [{ html }] }`. Because every article is rendered in a `Promise.all` and `renderBlock` throws on anything unexpected, a single renamed field took down the whole feed. `/theverge/apple` was additionally hitting `Unsupported block type: VideoBlockType` / `FeaturedProductsBlockType`.

**Changes.**
- Adapt to the renamed paragraph fields (backwards compatible with the old names).
- Fetch the feed via `ofetch` + `parser.parseString` instead of `parser.parseURL`, which relies on Node's `https.get` and is unavailable on Cloudflare Workers (`[unenv] https.get is not implemented yet!` on rsshub.app).
- Unknown or malformed blocks are skipped with a warning instead of throwing.
- Layered fallbacks per article: `__NEXT_DATA__` → server-rendered article body (`.duet--article--article-body-component`) → feed summary. The last one is not cached so transient fetch failures are retried.
- Render `VideoBlockType` (same Volume embed as the lede video); exclude `FeaturedProductsBlockType` like the other product-table blocks.
- `item.category` was set to the raw category objects (rendered as `[object Object]`); now uses their titles.
- Refresh the hub table: `gaming`, `hd`, `photography` no longer publish; list the hubs that currently do (verified against `theverge.com/rss/<hub>/index.xml`).

Verified locally on `/theverge`, `/theverge/apple`, `/theverge/games`, `/theverge/reviews` (10 items each, full content, no warnings), and exercised both fallback paths by temporarily breaking the `__NEXT_DATA__` selector and the article fetch.

### DIFF
--- a/lib/routes/theverge/index.ts
+++ b/lib/routes/theverge/index.ts
@@ -8,7 +8,7 @@ import parser from '@/utils/rss-parser';
 
 import { renderHeader } from './templates/header';
 
-const excludeTypes = new Set(['NewsletterBlockType', 'RelatedPostsBlockType', 'ProductsTableBlockType', 'FeaturedProductsBlockType', 'TableOfContentsBlockType']);
+const excludeTypes = new Set(['ActionBoxBlockType', 'NewsletterBlockType', 'RelatedPostsBlockType', 'ProductsTableBlockType', 'FeaturedProductsBlockType', 'TableOfContentsBlockType']);
 
 const shouldKeep = (b: any) => !excludeTypes.has(b.__typename);
 
@@ -16,16 +16,11 @@ const shouldKeep = (b: any) => !excludeTypes.has(b.__typename);
 const renderContents = (b: any): string => (b.paragraphContents ?? b.tempContents ?? [b.contents]).map((c) => c?.html ?? '').join('');
 
 // Render a list of blocks; a single unknown or malformed block is dropped with a warning instead of failing the whole article
+// Errors are deliberately not caught here: an unrenderable block makes the whole article fall back to the
+// server-rendered body, which is still full text, rather than silently emitting a description with a hole in it
 const renderBlocks = (blocks: any[] | undefined, separator: string): string =>
     (blocks ?? [])
-        .map((b) => {
-            try {
-                return renderBlock(b);
-            } catch (error) {
-                logger.warn(`theverge: failed to render ${b?.__typename} block: ${error}`);
-                return '';
-            }
-        })
+        .map((b) => renderBlock(b))
         .filter(Boolean)
         .join(separator);
 
@@ -126,8 +121,7 @@ const renderBlock = (b) => {
         case 'TableBlockType':
             return `<table><tr>${b.header.map((cell) => `<th>${cell}</th>`).join('')}</tr>${b.rows.map((row) => `<tr>${row.map((cell) => `<td>${cell}</td>`).join('')}</tr>`).join('')}</table>`;
         default:
-            logger.warn(`theverge: unsupported block type ${b.__typename}`);
-            return '';
+            throw new Error(`Unsupported block type: ${b.__typename}`);
     }
 };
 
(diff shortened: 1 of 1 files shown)


lib/routes/theverge/index.ts
```diff
@@ -32,20 +45,36 @@ export const route: Route = {
...
+| music          | Music          |
+| streaming      | Streaming      |
+| creators       | Creators       |
+| culture        | Culture        |
+| policy         | Policy         |
+| business       | Business       |
+| transportation | Transportation |
+| space          | Space          |
+| health         | Health         |
+| web            | Web            |
```
TonyRL, line 72: Keep the entries [sorted](https://docs.rsshub.app/joinus/advanced/script-standard) in alphabetical order.

lib/routes/theverge/index.ts
```diff
@@ -87,64 +116,98 @@ const renderBlock = (b) => {
...
+
 async function handler(ctx) {
     const link = ctx.req.param('hub') ? `https://www.theverge.com/rss/${ctx.req.param('hub')}/index.xml` : 'https://www.theverge.com/rss/index.xml';
 
-    const feed = await parser.parseURL(link);
+    // rss-parser's parseURL relies on Node's https.get, which is unavailable on Cloudflare Workers
+    const feedText = await ofetch(link, {
+        parseResponse: (text) => text,
+    });
+    const feed = await parser.parseString(feedText);
```
TonyRL: Unncessary changes. This isn't true at all. https://rsshub.app runs on Cloudflare Workers and routes like https://github.com/DIYgod/RSSHub/blob/d629c66a42faaa920a8658bb401fde2896e4653f/lib/routes/9to5/subsite.ts#L61 works completely fine.

lib/routes/theverge/index.ts
```diff
@@ -87,64 +116,98 @@ const renderBlock = (b) => {
...
+                        logger.warn(`theverge: failed to render ${item.link} from __NEXT_DATA__, falling back to page HTML: ${error}`);
+                        const body = $('.duet--article--article-body-component');
+                        if (body.length === 0) {
+                            throw new Error(`neither __NEXT_DATA__ nor the article body could be read from ${item.link}`, { cause: error });
+                        }
+                        item.description = body
+                            .toArray()
+                            .map((el) => $(el).html())
+                            .join('');
+                    }
```
TonyRL: Could you show me which article can not `node` and requires using `.duet--article--article-body-component`?

lib/routes/theverge/index.ts
```diff
@@ -87,64 +116,98 @@ const renderBlock = (b) => {
...
-                }
-
-                item.description = description;
-                item.category = node.categories;
-
+            } catch (error) {
+                // A single unreadable article should not take down the whole feed; this result is not cached, so it is retried next time
+                logger.warn(`theverge: falling back to the feed summary for ${item.link}: ${error}`);
+                // the shared rss-parser does not expose Atom <category> elements, so this path carries no categories
+                item.description = item.content;
```
TonyRL: Could you show me which article is unreadable and requires using the fallback `item.content`?

TonyRL (changes requested):

IvanWng97:
All four addressed in 9c88e85.

- Hub table sorted alphabetically.
- `parser.parseURL` restored — unrelated to this fix, agreed.
- Both fallbacks removed. I checked all 79 distinct articles across the 10 hub feeds and every one renders from `node`, so I had no case for either fallback.

What's left is the actual breakage: `tempContents` → `paragraphContents` (and stream excerpts from `{contents}` to `{paragraphContents}`), `VideoBlockType`, excluding `FeaturedProductsBlockType`, `item.category` using titles instead of the raw objects, and the hub list (`gaming`, `hd` and `photography` no longer publish).

lib/routes/theverge/index.ts
```diff
@@ -2,22 +2,18 @@ import { load } from 'cheerio';
 
 import type { DataItem, Route } from '@/types';
 import cache from '@/utils/cache';
-import logger from '@/utils/logger';
 import ofetch from '@/utils/ofetch';
 import parser from '@/utils/rss-parser';
 
 import { renderHeader } from './templates/header';
 
-const excludeTypes = new Set(['ActionBoxBlockType', 'NewsletterBlockType', 'RelatedPostsBlockType', 'ProductsTableBlockType', 'FeaturedProductsBlockType', 'TableOfContentsBlockType']);
```
TonyRL: Why remove `ActionBoxBlockType` from `excludeTypes`?

lib/routes/theverge/index.ts
```diff
@@ -2,22 +2,18 @@ import { load } from 'cheerio';
...
+const excludeTypes = new Set(['FeaturedProductsBlockType', 'NewsletterBlockType', 'ProductsTableBlockType', 'RelatedPostsBlockType', 'TableOfContentsBlockType']);
 
 const shouldKeep = (b: any) => !excludeTypes.has(b.__typename);
 
-// Rich text lives under `paragraphContents` (formerly `tempContents`), or as a single `contents` object on headings, list items and quotes
+// Rich text moved from `tempContents` to `paragraphContents`; headings, list items and quotes carry a single `contents`
 const renderContents = (b: any): string => (b.paragraphContents ?? b.tempContents ?? [b.contents]).map((c) => c?.html ?? '').join('');
```
TonyRL: Could you show me which article is still using `tempContents`?

IvanWng97:
Both fixed in 87ced44.

- `ActionBoxBlockType` — my mistake, it got dropped when I rebuilt this file from master in the previous commit. Restored; without it four of the hub feeds fail outright now that there's no per-article fallback.
- `tempContents` — no article behind it, removed.

PR #23083, opened 2026-08-24
feat(route): add 35mmc

IvanWng97:

N/A

```routes
/35mmc
/35mmc/5-frames-with
```

- [x] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

The official feed (`35mmc.com/feed`) only carries excerpts. This route reads the WordPress REST API and returns the full post with all images, featured image, author and categories, optionally filtered by category slug.

Rendering post bodies is slow on this site (~2s per post, and the host serialises concurrent requests), so the route only lists ids and fetches each post separately with `cache.tryGet`, so after the first run only new posts are fetched. Image `src` is forced to https (the site still stores plain-http URLs, which readers block as mixed content) and browser-only attributes (`srcset`, `loading`, …) are dropped. The site's users endpoint is disabled, so the author comes from the Yoast block.

### DIFF
--- a/lib/routes/35mmc/index.tsx
+++ b/lib/routes/35mmc/index.tsx
@@ -23,7 +23,7 @@ const api = (path: string, query: Record<string, string | number>) =>
 // srcset/lazy-loading attributes that only make sense in a browser
 const cleanContent = (html: string): string => {
     const $ = load(html, null, false);
-    $('img').each((_, el) => {
+    $('img, iframe').each((_, el) => {
         const $img = $(el);
         const src = $img.attr('src');
         if (src?.startsWith('http://')) {
--- a/lib/routes/35mmc/index.tsx
+++ b/lib/routes/35mmc/index.tsx
@@ -1,0 +1,151 @@
+import { load } from 'cheerio';
+import { decodeHTML } from 'entities';
+import { raw } from 'hono/html';
+import { renderToString } from 'hono/jsx/dom/server';
+
+import InvalidParameterError from '@/errors/types/invalid-parameter';
+import type { DataItem, Route } from '@/types';
+import { ViewType } from '@/types';
+import cache from '@/utils/cache';
+import ofetch from '@/utils/ofetch';
+import { parseDate } from '@/utils/parse-date';
+
+const baseUrl = 'https://www.35mmc.com';
+
+const api = (path: string, query: Record<string, string | number>) =>
+    ofetch(`${baseUrl}/wp-json/wp/v2/${path}`, {
+        query,
+        // the default browser-like Accept header can make WordPress serve the HTML page instead of JSON
+        headers: { accept: 'application/json' },
+    });
+
+// Post bodies still reference images over plain http (blocked as mixed content by readers) and carry
+// srcset/lazy-loading attributes that only make sense in a browser
+const cleanContent = (html: string): string => {
+    const $ = load(html, null, false);
+    $('img').each((_, el) => {
+        const $img = $(el);
+        const src = $img.attr('src');
+        if (src?.startsWith('http://')) {
+            $img.attr('src', src.replace(/^http:\/\//, 'https://'));
+        }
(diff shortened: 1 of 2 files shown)


IvanWng97:
@TonyRL all comments are resolved～ could u help take a look again? thanks

lib/routes/35mmc/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+import ofetch from '@/utils/ofetch';
+import { parseDate } from '@/utils/parse-date';
+
+const baseUrl = 'https://www.35mmc.com';
+
+const api = (path: string, query: Record<string, string | number>) =>
+    ofetch(`${baseUrl}/wp-json/wp/v2/${path}`, {
+        query,
+        // the default browser-like Accept header can make WordPress serve the HTML page instead of JSON
+        headers: { accept: 'application/json' },
```
TonyRL: Can you show me what the HTML page looks like when using the default browser-like accept headers?

lib/routes/35mmc/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+        : undefined;
+
+    // The posts endpoint renders every post body server-side (~2s per post), so only list ids here
+    // and fetch each post separately, in parallel and cached per post
+    const list = await api('posts', {
+        per_page: limit,
+        _fields: 'id,link',
+        ...(category && { categories: category.id }),
+    });
+    if (!Array.isArray(list)) {
```
TonyRL: `/wp-json/wp/v2/categories` and `/wp-json/wp/v2/posts` always return an array when the request is successful.
https://github.com/WordPress/wordpress-develop/blob/6c336a23a3765250a650f5066bbe96e04d058d17/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L459
https://github.com/WordPress/wordpress-develop/blob/6c336a23a3765250a650f5066bbe96e04d058d17/src/wp-includes/rest-api/endpoints/class-wp-rest-terms-controller.php#L366

lib/routes/35mmc/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+        throw new TypeError(`Unexpected response from the posts API: ${JSON.stringify(list).slice(0, 200)}`);
+    }
+
+    const items = await Promise.all(
+        list.map((entry) =>
```
TonyRL: Do not spam the remote server with multiple concurrent requests when a single request is sufficient.

lib/routes/35mmc/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+// srcset/lazy-loading attributes that only make sense in a browser
+const cleanContent = (html: string): string => {
+    const $ = load(html, null, false);
+    $('img, iframe').each((_, el) => {
+        const $img = $(el);
+        const src = $img.attr('src');
+        if (src?.startsWith('http://')) {
+            $img.attr('src', src.replace(/^http:\/\//, 'https://'));
+        }
+        $img.removeAttr('srcset').removeAttr('sizes').removeAttr('loading').removeAttr('decoding');
```
TonyRL: Why remove attributes like `loading="lazy"` and `decoding="async"`?

TonyRL (changes requested):

IvanWng97:
All four in 6a7b5d9.

- Single embedded request instead of listing ids and fetching each post. Default limit is now 10, which keeps that one request around 17s; at 20 it was 53s, which is what pushed me into the per-post fetches.
- `Array.isArray` guards dropped — you're right, those endpoints always answer with an array on success.
- `accept: application/json` dropped. I checked and this host returns `200 application/json` with the browser-like headers too, so I had no case for it.
- Image handling is now just the http to https rewrite, which this site does need — every image in the post bodies is referenced over plain http and readers block them as mixed content. `loading`, `decoding` and `srcset` are left as they come.

PR #23121, opened 2026-08-28
fix(route/latepost): include cover image and abstract in description

esonwong:

Close #

```routes
/latepost
/latepost/1
/latepost/2
/latepost/3
/latepost/4
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

This is a fix to an existing route, not a new one — the checklist above is left blank on purpose.

### What was wrong

`item.description` is built from `#select-main`, which is the article body only. The cover image sits *outside* it, in a sibling block, so feed items never included it even though nearly every recent article has one.

### The two layouts

The site uses different markup depending on the column:

| Column | Cover markup |
| --- | --- |
| 晚点独家 / 长报道 (`proma` 1, 4) | `` inside `.abstract` |
| 人物访谈 (`proma` 2) | `` |
| 晚点早知道 (`proma` 3) | no cover at all |

Handling only the first selector would have missed every interview, so the patch tries both and then falls back to the `var imgUrl` share image declared in the page script, which is present exactly when a cover exists and absent otherwise.

### Scheme upgrade

The cover is served as `http://www.latepost.com/uploads/cover/...` while the body images use https. Readers rendering the item over https may refuse the mixed content, so the scheme is upgraded. The body images are unaffected.

### Abstract

The one-line abstract shown next to the cover (`.abstract-pic-left`) was dropped as well; it is now prepended as a `<blockquote>`.

### Verification

Checked against saved copies of 11 real articles spanning all four columns:

- 晚点独家 / 长报道 — cover found via `#abstract_img`
- 人物访谈 — cover found via `img.interview-pic`
- 晚点早知道 — no cover in the source, correctly left empty rather than emitting a broken ``

Articles that genuinely have no cover produce the same description as before.

### DIFF
--- a/lib/routes/latepost/index.ts
+++ b/lib/routes/latepost/index.ts
@@ -94,7 +94,21 @@ async function handler(ctx): Promise<Data> {
                 const content = load(detailResponse);
 
                 item.title ??= content('div.article-header-title').text();
-                item.description = content('#select-main').html()!.replaceAll('<p><br></p>', '');
+
+                // The cover image lives outside of the article body (#select-main), in either the
+                // `.abstract` block (exclusives / long reads) or `.interview-pic` (interviews).
+                // Fall back to the share image embedded in the page script, which covers both layouts.
+                let cover = content('#abstract_img').attr('src') || content('img.interview-pic').attr('src');
+                if (!cover) {
+                    cover = /var\s+imgUrl\s*=\s*'([^']+)'/.exec(detailResponse)?.[1];
+                }
+                // The site serves the cover over plain http, which readers may refuse to load.
+                const coverHtml = cover ? `<img src="${new URL(cover, rootUrl).href.replace(/^http:/, 'https:')}">` : '';
+
+                const abstract = content('div.abstract-pic-left').text().trim();
+                const abstractHtml = abstract ? `<blockquote>${abstract}</blockquote>` : '';
+
+                item.description = coverHtml + abstractHtml + content('#select-main').html()!.replaceAll('<p><br></p>', '');
                 item.author = content('div.article-header-author div.author-link a.label').first().text();
                 item.category = item.category.filter(Boolean);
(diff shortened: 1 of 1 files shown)


lib/routes/latepost/index.ts
```diff
@@ -94,7 +94,21 @@ async function handler(ctx): Promise<Data> {
...
+                // `.abstract` block (exclusives / long reads) or `.interview-pic` (interviews).
+                // Fall back to the share image embedded in the page script, which covers both layouts.
+                let cover = content('#abstract_img').attr('src') || content('img.interview-pic').attr('src');
+                if (!cover) {
+                    cover = /var\s+imgUrl\s*=\s*'([^']+)'/.exec(detailResponse)?.[1];
+                }
+                // The site serves the cover over plain http, which readers may refuse to load.
+                const coverHtml = cover ? `<img src="${new URL(cover, rootUrl).href.replace(/^http:/, 'https:')}">` : '';
+
+                const abstract = content('div.abstract-pic-left').text().trim();
```
TonyRL: Can you show me which article's `div.abstract-pic-left` `text()` contains leading/trailing whitespaces that requires `trim()`?

TonyRL (changes requested):

lib/routes/latepost/index.ts
esonwong: None, actually. I checked the 50 articles with an abstract between id 3680 and 3741, and none of them has leading or trailing whitespace in `div.abstract-pic-left`, so the `trim()` was unnecessary. Removed it in 8e1d761.

I also switched `.text()` to `.html()` there, so the abstract stays entity-escaped when it is put inside the `<blockquote>` (the auto review pointed out the injection risk). The markup is plain text in every sample, so the output is the same.

PR #23126, opened 2026-08-28
fix(route/wikipedia): escape link targets in current events

aavanian:

Close #

```routes
/wikipedia/current-events
```

- [ ] New Route / 新的路由
    - [x] Follows Script Standard / 跟随路由规范
- [ ] Date and time / 日期和时间
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Fix to existing route.

  - [x] link targets are interpolated into the `href` attribute unescaped:
    - targets come from the wikitext, so any editor controls them, and `sanitizeHtml` runs on the wikitext before the tags are built and does not escape double quotes
    - `[[foo" onmouseover="alert(1)|x]]` produced an anchor carrying an `onmouseover` handler, and `[javascript:alert(1) x]` a `javascript:` link
    - fixed by building the url through `encodeURI` and only linkifying external urls when they are `http(s)`
  - [x] wiki hrefs contained raw spaces, making them invalid: fixed by the same encoding

### DIFF
--- a/lib/routes/wikipedia/current-events.ts
+++ b/lib/routes/wikipedia/current-events.ts
@@ -95,17 +95,43 @@ function stripTemplates(wikitext: string): string {
     return wikitext.replaceAll(/\{\{([^}]+)\}\}/g, '$1');
 }
 
+// sanitizeHtml runs on the wikitext before any tag is built, so & and < reach this point already
+// encoded. Decode them before assembling a URL, or the entity text ends up inside the link. &amp;
+// must go last so that &amp;lt; does not decode all the way to <.
+function decodeSanitizedEntities(text: string): string {
+    return text.replaceAll('&lt;', '<').replaceAll('&gt;', '>').replaceAll('&quot;', '"').replaceAll('&amp;', '&');
+}
+
+// Link targets come from the wikitext, so they are attacker-controlled and land in an href
+// attribute. encodeURI percent-encodes the " that would otherwise close the attribute and let the
+// rest of a crafted target through as markup, and it fixes the spaces that made the URLs invalid.
+function wikiUrl(target: string): string {
+    return `https://en.wikipedia.org/wiki/${encodeURI(decodeSanitizedEntities(target).trim().replaceAll(' ', '_'))}`;
+}
+
+// Anything that is not http(s) — javascript:, data: — is left as plain text rather than linked.
+function externalUrl(url: string): string | null {
+    const decoded = decodeSanitizedEntities(url).trim();
+    return /^https?:\/\//i.test(decoded) ? encodeURI(decoded) : null;
+}
+
 function convertWikiLinks(html: string): string {
     // Convert wiki links [[Link|Text]] or [[Link]]
-    html = html.replaceAll(/\[\[([^|\]]+)\|([^\]]+)\]\]/g, '<a href="https://en.wikipedia.org/wiki/$1">$2</a>');
-    html = html.replaceAll(/\[\[([^\]]+)\]\]/g, '<a href="https://en.wikipedia.org/wiki/$1">$1</a>');
+    html = html.replaceAll(/\[\[([^|\]]+)\|([^\]]+)\]\]/g, (_match, target: string, text: string) => `<a href="${wikiUrl(target)}">${text}</a>`);
+    html = html.replaceAll(/\[\[([^\]]+)\]\]/g, (_match, target: string) => `<a href="${wikiUrl(target)}">${target}</a>`);
     return html;
 }
 
 function convertExternalLinks(html: string): string {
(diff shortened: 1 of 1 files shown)


aavanian:
I believe this is ready to merge. let me know if anything is needed on my side.

lib/routes/wikipedia/current-events.ts
```diff
@@ -95,17 +95,43 @@ function stripTemplates(wikitext: string): string {
     return wikitext.replaceAll(/\{\{([^}]+)\}\}/g, '$1');
 }
 
+// sanitizeHtml runs on the wikitext before any tag is built, so & and < reach this point already
+// encoded. Decode them before assembling a URL, or the entity text ends up inside the link. &amp;
+// must go last so that &amp;lt; does not decode all the way to <.
+function decodeSanitizedEntities(text: string): string {
+    return text.replaceAll('&lt;', '<').replaceAll('&gt;', '>').replaceAll('&quot;', '"').replaceAll('&amp;', '&');
```
TonyRL: Use `entities` instead of manually replacing.

lib/routes/wikipedia/current-events.ts
```diff
@@ -95,17 +95,43 @@ function stripTemplates(wikitext: string): string {
...
+// Link targets come from the wikitext, so they are attacker-controlled and land in an href
+// attribute. encodeURI percent-encodes the " that would otherwise close the attribute and let the
+// rest of a crafted target through as markup, and it fixes the spaces that made the URLs invalid.
+function wikiUrl(target: string): string {
+    return `https://en.wikipedia.org/wiki/${encodeURI(decodeSanitizedEntities(target).trim().replaceAll(' ', '_'))}`;
+}
+
+// Anything that is not http(s) — javascript:, data: — is left as plain text rather than linked.
+function externalUrl(url: string): string | null {
+    const decoded = decodeSanitizedEntities(url).trim();
```
TonyRL: Can you show me what can be `trim()`?

lib/routes/wikipedia/current-events.ts
aavanian: Thanks, simpler and cleaner indeed. Fixed in 7800f49

lib/routes/wikipedia/current-events.ts
aavanian: this `trim()` was indeed unnecessary with the custom `decodeSanitizedEntities` but with `entities`'s `decodeXML` (7800f49), spaces could be re-introduced so I left it in place.

TonyRL: Can you show me how spaces can be re-introduced?

aavanian:
```
import { decodeXML } from 'entities';

const HTTP_RE = /^https?:\/\//i;

function withoutTrim(capturedUrl) {
    const decoded = decodeXML(capturedUrl);
    return HTTP_RE.test(decoded) ? encodeURI(decoded) : null;
}

function withTrim(capturedUrl) {
    const decoded = decodeXML(capturedUrl).trim();
    return HTTP_RE.test(decoded) ? encodeURI(decoded) : null;
}

const cases = [
    '&#32;http://example.com',   // decimal ref for U+0020 SPACE, prepended
    'http://example.com&#9;',    // decimal ref for U+0009 TAB, appended
];

for (const capturedUrl of cases) {
    console.log('captured url :', JSON.stringify(capturedUrl));
    console.log('decoded      :', JSON.stringify(decodeXML(capturedUrl)));
    console.log('without trim :', withoutTrim(capturedUrl));
    console.log('with trim    :', withTrim(capturedUrl));
    console.log('---');
}
```

TonyRL: All the inputs got https://github.com/DIYgod/RSSHub/blob/7800f49f08f5f91c6b9eb6104242ef7907608010/lib/routes/wikipedia/current-events.ts#L273 which is simply https://github.com/DIYgod/RSSHub/blob/7800f49f08f5f91c6b9eb6104242ef7907608010/lib/routes/wikipedia/current-events.ts#L265 Do those values *survive* `sanitize-html`?

It then got passed to https://github.com/DIYgod/RSSHub/blob/7800f49f08f5f91c6b9eb6104242ef7907608010/lib/routes/wikipedia/current-events.ts#L275 https://github.com/DIYgod/RSSHub/blob/7800f49f08f5f91c6b9eb6104242ef7907608010/lib/routes/wikipedia/current-events.ts#L121-L128 Do those values really match with those regular expressions?
Do you really know the logic of your code?

aavanian: indeed. i only tested the difference between decodeXML and the previous decodeSanitizedEntities

Apologies for the time wasted

PR #23131, opened 2026-08-29
fix(route/bilibili): guard against null emoji_details in followings dynamic

### LINKED ISSUE #21916: [Bug] /bilibili/followings/dynamic showEmoji=1 导致 TypeError: t is not iterable
### 路由地址

```routes
/bilibili/followings/dynamic/:uid/:routeParams?
```

### 完整路由地址

```fullroutes
/bilibili/followings/dynamic/:uid/showEmoji=1
```

### 相关文档

https://docs.rsshub.app/routes/social-media#bilibili

### 预期是什么？

正常返回关注动态 RSS，并渲染表情图片

### 实际发生了什么？

返回 503，日志：TypeError: t is not iterable

### 部署

自建

### 部署相关信息

Docker diygod/rsshub:latest, Ubuntu 22.04, Node v24.15.0, 已配置 BILIBILI_COOKIE

### 额外信息

```shell
根因：followings-dynamic.ts 中 emoji_details 未判空。修复：item.display?.emoji_info?.emoji_details && showEmoji

本bug为我使用ai工具部署rsshub时大模型发现并反应，bug报告也是由大模型生成
```

### 这不是重复的 issue

- [x] 我已经搜索了 [现有 issue](https://github.com/DIYgod/RSSHub/issues)，以确保该错误尚未被报告。

ryanchou1994:

Close #21916

```routes
/bilibili/followings/dynamic/109937383/showEmoji=1
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

Not a new route — a one-line fix to an existing one, so the checklist does not apply.

## Note / 说明

The emoji guard checks `emoji_info`, but the loop iterates `emoji_info.emoji_details`. For dynamics carrying an `emoji_info` object with no emoji, `emoji_details` is `null`, so `for...of` throws `TypeError: t is not iterable` and the route 503s whenever `showEmoji=1`. The reporter of #21916 pinned it to this line; #21914 and #21915 are the same crash.

The guard now checks the value that is actually iterated, and `showEmoji` moved first so the cheap check runs first.

Minimal repro of both guards (Node 26) over the shapes the API returns:

| `item.display.emoji_info` | before | after |
|---|---|---|
| `{ emoji_details: null }` | `TypeError` | skipped |
| `{}` (field absent) | `TypeError` | skipped |
| `{ emoji_details: [1 emoji] }` | 1 emoji rendered | 1 emoji rendered |
| no `display` | skipped | skipped |

Crashing shapes are handled, normal rendering is unchanged, and `emoji_details` appears nowhere else in the repo.

Replaces #23130: my first description dropped two required headings, so the route detector could not read the `routes` block and auto-closed it. Same commit, corrected description.

### DIFF
--- a/lib/routes/bilibili/followings-dynamic.ts
+++ b/lib/routes/bilibili/followings-dynamic.ts
@@ -174,7 +174,7 @@ async function handler(ctx) {
 
             // emoji
             let data_content = getDes(data);
-            if (item.display && item.display.emoji_info && showEmoji) {
+            if (showEmoji && item.display?.emoji_info?.emoji_details) {
                 const emoji = item.display.emoji_info.emoji_details;
                 for (const item of emoji) {
                     data_content = data_content.replaceAll(


ryanchou1994:
The route test failure above is `ConfigNotFoundError: 缺少对应 uid 的 Bilibili 用户登录后的 Cookie 值` — `/bilibili/followings/dynamic` needs a `BILIBILI_COOKIE_<uid>`, which CI doesn't have, so this route reports Failed no matter what the diff does.

The change doesn't touch the request path: the crash happens after the response is parsed, when the emoji block iterates `emoji_info.emoji_details` while only having checked `emoji_info`. I verified it by running both guards over the shapes the API returns (table in the description) rather than through the live route.

Let me know if you'd rather I put a different, CI-testable route in the `routes` block.

PR #23143, opened 2026-08-30
feat(route/nytimes): add Wirecutter

IvanWng97:

N/A

```routes
/nytimes/wirecutter
```

- [x] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

**Why.** Wirecutter has an official feed at `nytimes.com/wirecutter/feed/`, but it only carries the opening few paragraphs. For *The Best Plain Yogurt* that is ~4KB of `content:encoded` against a ~29KB review — the feed stops before any of the actual picks. Wirecutter reviews are also structured as chapters with product callouts, which is exactly the part that gets cut.

**How.** The feed supplies the item list; each review is then rendered from `props.pageProps.post` in the page's `__NEXT_DATA__`:

- chapter titles become headings, and each chapter's `body` node tree is rendered through a tag/attribute allowlist (layout classes and `data-gtm-*` tracking attributes are dropped)
- `shortcode-gallery` becomes figures with the photo credit as the caption, `shortcode-caption` becomes a `figure`/`figcaption`, and `shortcode-callout` — the picks — renders as the ribbon, product name, photo, verdict and description
- `adslot`, `shortcode-recirc` and `shortcode-scoop_form_callout` are dropped as ads, in-house promos and a form
- images drop the resizing query string so the original is served

**Failure handling.** An unknown node type throws rather than being silently skipped, so that review falls back to the feed summary instead of emitting a body with a hole in it. That fallback happens outside `cache.tryGet`, so it is not cached and is retried on the next run; one unreadable review never takes down the feed. Verified by breaking the `__NEXT_DATA__` selector: all 10 items degrade to the summary and the route still returns 200.

The node inventory was taken across 8 reviews (1,300+ nodes) to make sure the allowlist and the dropped types cover what Wirecutter actually publishes.

### DIFF
--- a/lib/routes/nytimes/templates/wirecutter.tsx
+++ b/lib/routes/nytimes/templates/wirecutter.tsx
@@ -1,0 +1,156 @@
+import type { FC } from 'hono/jsx';
+import { renderToString } from 'hono/jsx/dom/server';
+
+const imageBase = 'https://cdn.thewirecutter.com/';
+
+// Presentational or promotional nodes that carry no article content
+const dropTypes = new Set(['adslot', 'shortcode-recirc', 'shortcode-scoop_form_callout']);
+
+// hono/jsx only applies void-element rules to literal tags, so a dynamic <Tag> must be self-closed
+// explicitly: <Tag></Tag> would emit `<br></br>`, which HTML parsers turn into two line breaks
+const voidTags = new Set(['br', 'hr', 'img', 'source']);
+
+const allowedTags = new Set([
+    'p',
+    'a',
+    'strong',
+    'b',
+    'em',
+    'i',
+    'h2',
+    'h3',
+    'h4',
+    'h5',
+    'h6',
+    'ul',
+    'ol',
+    'li',
+    'br',
+    'hr',
+    'img',
+    'blockquote',
+    'figure',
+    'figcaption',
+    'div',
+    'span',
+    'video',
+    'source',
+    'table',
+    'thead',
+    'tbody',
+    'tr',
+    'th',
+    'td',
+]);
+// Everything else is layout plumbing or tracking metadata
+const allowedAttributes = new Set(['href', 'src', 'alt', 'title', 'width', 'height', 'colspan', 'rowspan']);
+
+// Images are served through a resizing CDN; without the query string the original is returned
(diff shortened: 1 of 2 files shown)


IvanWng97:
Both auto-review findings are addressed in 281f992.

**`parseDate`** — applied at both call sites, as `item.pubDate ? parseDate(item.pubDate) : undefined` following `lib/routes/baoyu/index.ts:47`, so that a missing date stays undefined rather than becoming a fake one.

**JSX rendering** — the body renderer moved to `templates/wirecutter.tsx` and is now built from JSX components, so escaping comes from the renderer and `escapeText` / `escapeAttribute` are gone. It sits in `templates/` rather than renaming the route file, matching `scientificamerican/templates/description.tsx`, which is the closest existing case (a node tree rendered through `const Tag = ... as unknown as FC`).

One part of that suggestion does not hold, so noting it here for the next reader: **the void-tag set cannot be dropped.** `hono/jsx` only applies void-element rules to literal tags, not to a dynamic `<Tag>`:

```
static  <br />        ->  "<br/>"
dynamic Tag = 'br'    ->  "<br></br>"
dynamic Tag = 'img'   ->  "<img></img>"
dynamic Tag = 'hr'    ->  "<hr></hr>"
```

HTML parsers turn a stray `</br>` into a second line break, so the first version of this refactor rendered two source `br` nodes as four. Void tags are therefore self-closed explicitly (`<Tag {...attributes} />`) with a comment recording why.

With that in place the rendered `content_html` is byte-identical to the previous string-based renderer across all 10 items in the feed, and breaking the `__NEXT_DATA__` selector still degrades every item to the feed summary with a 200 response.

IvanWng97:
d06c0d2 should fix the route test failure.

The article pages sit behind bot detection that turns away a share of requests with a 403 — from GitHub's runners as well, which is what the earlier `Failed ❌` was. `403` is not in ofetch's default `retryStatusCodes`, so one refusal took the whole feed down. Retrying clears it: measured over the feed, roughly half the articles are refused on a first pass and every one of them answers on a later attempt, so three cold runs now return all ten articles with full text, using 6 to 13 retries between them.

No fallback involved — a render error still surfaces. That also caught the last unhandled block type, `shortcode-pullquote`, which is implemented here too. Route is marked `antiCrawler`.

PR #23208, opened 2026-09-05
feat(route/mikiki): add mikiki latest articles route

ashi-koki:

Close #

```routes
/mikiki
```

- [x] New Route / 新的路由
    - [x] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Adds a route for MIKIKI (ミュージック・マガジン増刊, mikiki.tokyo.jp), a Japanese music magazine.

- Native RSS feed (`/list/feed/rss`) is fetched via `ofetch` (retry-capable), then each item's article page is fetched to extract the full article body, author and hero image.
- Article content is scraped with `cheerio` from `.article-body`, removing the table-of-contents (`.article-index`) and in-article ranking widgets (`.ranking-in-article`).
- Author is read from `.article-header-author-name`; publish date is taken directly from the feed's `dc:date` (JST).
- Per-item results are cached with `cache.tryGet`.

### DIFF
--- a/lib/routes/mikiki/index.ts
+++ b/lib/routes/mikiki/index.ts
@@ -3,6 +3,7 @@ import { load } from 'cheerio';
 import type { Data, DataItem, Route } from '@/types';
 import cache from '@/utils/cache';
 import ofetch from '@/utils/ofetch';
+import { parseDate } from '@/utils/parse-date';
 import parser from '@/utils/rss-parser';
 
 export const route: Route = {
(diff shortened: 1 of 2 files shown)


lib/routes/mikiki/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    example: '/mikiki',
+    name: '最新記事',
+    maintainers: ['ashi-koki'],
+    handler,
+};
+
+async function handler(): Promise<Data> {
+    const rootUrl = 'https://mikiki.tokyo.jp';
+
+    const feedResponse = await ofetch(`${rootUrl}/list/feed/rss`, { responseType: 'text' });
```
TonyRL: What specify the [default option](https://github.com/unjs/ofetch/blob/1dbc37fd1ceab832fc7c90cad81b1091c95ba563/src/utils.ts#L78) again? Does the feed break without it?

ashi-koki: Right — the feed returns `Content-Type: text/xml`, which ofetch's `detectResponseType` already treats as `text` (via the `text/*` prefix rule), so `responseType: 'text'` was redundant. Removed it.

PR #23301, opened 2026-09-15
fix(route): gate the user supplied hosts in discuz and yna

kah-ja:

Close #

```routes
/discuz/x/https%3a%2f%2fwww.52pojie.cn%2fforum-16-1.html
/yna/en/national
/yna/ko/economy
```

- [ ] New Route / 新的路由
    - [ ] Parsed / 可以解析
- [ ] `Puppeteer`

## Note / 说明

Two existing routes decide the request host from path parameters without checking them.

### `/discuz/:link`

`lib/routes/discuz/discuz.ts:98` takes the whole trailing path as the URL and hands it to `fetchWithAntiBot` -> `ofetch.raw` (`:112`, `:26`), so the deployment fetches whatever host the caller names, loopback and link local addresses included, and reflects the title, description and first post back into the feed.

This is the same situation the project already handles with `ALLOW_USER_SUPPLY_UNSAFE_DOMAIN` in `lib/routes/wordpress/index.ts:16`, `lib/routes/hexo/fluid.ts:25` and the mastodon and misskey routes. The route now uses that gate, declares it in `features.requireConfig` so it shows up in the docs, and reuses the `^https?://...` shape check from the wordpress route.

**Behaviour change:** on a deployment without `ALLOW_USER_SUPPLY_UNSAFE_DOMAIN=true`, `/discuz/...` now returns the same "disabled unless ..." error as the other generic routes instead of fetching. If you would rather keep the route open by default and only block private targets, say so and I will rework it that way.

### `/yna/:lang?/:channel?`

`lib/routes/yna/index.ts:59` interpolates `lang` as the leading label of the host, so `lang=127.0.0.1:8080%23` moves the request off `yna.co.kr`. The handler now accepts only the codes the route documents (`ko en cn jp ar es fr`), plus `sp`, because the Spanish feed page linked in the route description lives on `sp.yna.co.kr`, and throws `InvalidParameterError` otherwise. `channel` is left as it is: it only selects a path on the same host.

### Testing

- `pnpm typecheck` clean, `pnpm exec oxlint --type-aware --config=.oxlintrc.ci.json lib/routes/discuz lib/routes/yna` reports 0 warnings and 0 errors, `oxfmt --check` clean on both files.
- No route test exists for either route and none was added; the changes are input validation before the request is made.
- The full vitest suite was not run.

### References

GHSA-7mqg-69mf-j9vm (reported privately, unpublished at time of writing)

Found during a [penetration test](https://www.turingpoint.de/en/security-assessments/pentests/) by turingpoint.


kah-ja:
The forks the bot flagged are the fix branches for a series of privately reported security issues (this one is GHSA-7mqg-69mf-j9vm) and the change itself is input validation in two routes, so could you reopen?

